PR #4, opened 2024-07-25
INFRA - Initialise Core

kshxtij:
Begin core implementation in `lib/core` to port over the contents from hyped-2024.

**Minimum Viable Core:**

- [x] MQTT client
- [x] MQTT topics
- [x] Logging
- [x] Include root `Cargo.toml` file and `.gitignore`


kshxtij:
Fwiw on this PR, @davidbeechey can we make a wishlist for the "minimum viable core" since most future PRs will depend and update this anyway.

lib/core/src/lib.rs
```diff
@@ -0,0 +1,6 @@
+#![cfg_attr(not(feature = "std"), no_std)]
+
+pub mod format_string;
+pub mod logger;
```
davidbeechey:
```suggestion
pub mod log_types;
```

lib/core/src/mqtt_topics.rs
davidbeechey, line 1: It's a bit annoying having an `std` and `no_std` version of `to_string` for `MqttTopics`. Is there a way round it?

lib/core/src/mqtt.rs
```diff
@@ -0,0 +1,122 @@
...
+}
+
+#[cfg(feature = "std")]
+pub struct MqttMessage {
+    pub topic: String,
+    pub payload: String,
+}
+
+#[derive(Serialize, Deserialize)]
+pub struct ButtonMqttMessage {
```
davidbeechey: I think this can be removed?

lib/core/src/mqtt.rs
```diff
@@ -0,0 +1,122 @@
...
+    T: embedded_io_async::Read + embedded_io_async::Write,
+    R: rand_core::RngCore,
+> {
+    pub client: MqttClient<'a, T, 5, R>,
+}
+
+pub fn initialise_mqtt_config() -> ClientConfig<'static, 5, CountingRng> {
+    let mut config = ClientConfig::new(
+        rust_mqtt::client::client_config::MqttVersion::MQTTv5,
+        CountingRng(20000),
```
davidbeechey, line 51: Why 20000?

lib/core/src/mqtt.rs
```diff
@@ -0,0 +1,122 @@
...
+    );
+    config.add_max_subscribe_qos(rust_mqtt::packet::v5::publish_packet::QualityOfService::QoS1);
```
davidbeechey, line 53: I can't remember if you said this supports QoS2 or not?

lib/core/src/mqtt.rs
```diff
@@ -0,0 +1,122 @@
...
+    config.add_client_id("stm-client");
+    config.max_packet_size = 100;
+
+    config
+}
+
+// Implement send_message for HypedMqttClient
+impl<'a, T: embedded_io_async::Read + embedded_io_async::Write, R: rand_core::RngCore>
```
davidbeechey, line 59: Do we want a `new` function here which handles the config rather than the `initialise_mqtt_config` function (means `client` doesn't need to be public too I think)? Could pass in things like the broker address and client ID too

davidbeechey (commented):
Looks good, I think this should be all we need to get `core` started. Could we add the root `Config.toml` and `.gitignore` to this PR too so we have the structure set up in this PR too?

lib/core/src/mqtt_topics.rs
kshxtij: Yeah we can just use heapless universallly

lib/core/src/mqtt.rs
kshxtij: Complete honesty, idk. It's the example for the library and the number is just used as the starting value for the counter to emit nonces.

lib/core/src/mqtt.rs
kshxtij: Only supports QoS 1 currently, i was going to modify a fork of the library at some point to add QoS2

lib/core/src/mqtt.rs
kshxtij: Broker address happens at the socket level not the mqtt level so wouldn't work without a lot of borrow checker crying i think but beyond that this has been changed.

lib/core/src/mqtt.rs
davidbeechey: (We've just had a discussion about this on Slack, QoS2 not needed in favour of manual acks.)

davidbeechey (approved):

kshxtij:
Ok, im not actually sure what's wrong with the generics in the Mqtt::new() but i'd appreciate it if someone could also have a look.

Cargo.toml
```diff
@@ -1,3 +1,5 @@
 [workspace]
-members = []
+members = [
+    "lib/core"
```
davidbeechey:
```suggestion
    "lib/*"
```

I think this is valid? Just so we can add all the other `lib/` crates later without defining them all separately

davidbeechey (approved):
Looks good, happy to merge this once @TomLonergan03 has had a look too

Cargo.toml
kshxtij: Oh yeah oops, I think this is just from when I made my own file in here before it was in main.

PR #5, opened 2024-08-03
INFRA - Telemetry Board (HYPE-17)

kshxtij:
This pr initialises the code for the Telemetry board (as of now) which connects to the MQTT broker running on the base station and acts as a intermediary for all can messages and mqtt messages between the rest of the pod and the base station.

This PR currently contains boiler plate code to stress test the connection including a button based interrupt data sender as well as a time based task. It also handles receiving messages from the broker on subscribed topics.

This PR also includes an initial idea on how to handle on-board logging using defmt but also transmitting these logs over mqtt which will need to be further extended for the eventual can-based logs. 

sidenote: idk if infra is the tag you want on the pr


boards/stm32h743zi/src/bin/telemetry.rs
```diff
@@ -0,0 +1,125 @@
...
+        p.PG13,
+        p.PB13,
+        p.PG11,
+        GenericSMI::new(0),
+        mac_addr,
+    );
+
+    // let config = embassy_net::Config::dhcpv4(Default::default());
+    let config = embassy_net::Config::ipv4_static(embassy_net::StaticConfigV4 {
+        address: Ipv4Cidr::new(Ipv4Address::new(169, 254, 195, 61), 24),
```
davidbeechey: These addresses should be in the config file

boards/stm32h743zi/src/tasks/mqtt_send_and_recv.rs
```diff
@@ -0,0 +1,139 @@
...
+/// Task for receiving messages from the MQTT broker
+#[embassy_executor::task]
+pub async fn mqtt_recv_task(stack: &'static Stack<Ethernet<'static, ETH, GenericSMI>>) {
+    let mut rx_buffer: [u8; 4096] = [0; 4096];
+    let mut tx_buffer: [u8; 4096] = [0; 4096];
+    let mut socket = TcpSocket::new(&stack, &mut rx_buffer, &mut tx_buffer);
+    socket.set_timeout(Some(embassy_time::Duration::from_secs(600)));
+    log(LogLevel::Info, "Connecting to Receive Socket...").await;
+    match socket
+        .connect((Ipv4Address::new(169, 254, 195, 141), 1883))
```
davidbeechey: Likewise with the broker addresses

davidbeechey (changes requested):
I've tested this by plugging the board into my router and setting the IPs for my gateway and laptop accordingly, and it worked perfectly.

All this needs now is moving all config into `lib/config`, tidying/documenting a wee bit, and deciding whether we keep the test tasks in there or not (we could have a separate bin file for running it with tests?)

davidbeechey (changes requested):
Also please change the topics to some new test topics you can add to `hyped_core`

boards/stm32h743zi/src/tasks/mqtt_send_and_recv.rs
```diff
@@ -0,0 +1,140 @@
...
+/// Task for receiving messages from the MQTT broker
+#[embassy_executor::task]
+pub async fn mqtt_recv_task(stack: &'static Stack<Ethernet<'static, ETH, GenericSMI>>) {
+    let mut rx_buffer: [u8; 4096] = [0; 4096];
+    let mut tx_buffer: [u8; 4096] = [0; 4096];
+    let mut socket = TcpSocket::new(&stack, &mut rx_buffer, &mut tx_buffer);
+    socket.set_timeout(Some(embassy_time::Duration::from_secs(600)));
+    log(LogLevel::Info, "Connecting to Receive Socket...").await;
+    match socket
+        .connect(GATEWAY_ADDRESS)
```
davidbeechey: This should be called `MQTT_BROKER_ADDRESS`. I think the way Kshitij had networking configured meant that the same IP was used for both

boards/stm32h743zi/src/bin/telemetry.rs
```diff
@@ -0,0 +1,127 @@
...
+        p.PG13,
+        p.PB13,
+        p.PG11,
+        GenericSMI::new(0),
+        mac_addr,
+    );
+
+    // let config = embassy_net::Config::dhcpv4(Default::default());
+    let config = embassy_net::Config::ipv4_static(embassy_net::StaticConfigV4 {
+        address: BROKER_CIDR,
```
davidbeechey: `BROKER_CIDR` I think is actually the static IP that the board will have, so should be called `BOARD_STATIC_IP` or something instead

davidbeechey (changes requested):

davidbeechey (approved):
Looks good, obviously the main function will be changing once CAN is finished to code that will relay messages between CAN and MQTT. Until then I'm happy having the test code in there instead

platinumxy (approved):

boards/stm32h743zi/src/bin/telemetry.rs
```diff
@@ -0,0 +1,127 @@
...
+        });
+        config.rcc.pll1 = Some(rcc::Pll {
+            source: PllSource::HSE,
+            prediv: PllPreDiv::DIV4,
+            mul: PllMul::MUL216,
+            divp: Some(rcc::PllDiv::DIV2),
+            divq: None,
+            divr: None,
+        });
+        config.rcc.ahb_pre = AHBPrescaler::DIV1;
```
kshxtij: https://github.com/embassy-rs/embassy/blob/f319f1bc1b36cd6c7860391e49083ee64c079547/examples/stm32h7/src/bin/eth.rs#L33 ?? Not sure about each individual clock value

kshxtij (commented):
LGTM, just unsure about the clock values but we can swap that if we run into problems to match the stm32h7 example from embassy-rs

PR #6, opened 2024-08-10
SNS - Keyence (HYPE-12)

davidbeechey:
Keyence implementation in Rust for STM32.

~**New crates:**
I've created two new crates: `hyped_io` for IO abstractions and `hyped_sensors` for sensor implementations. (Both of these to use the `no_std` environment.)~

**GPIO implementation:**
To implement `Keyence` I've written the first IO code for GPIO. I'm suggesting the following structure (I am open to suggestions for other ways to do this!):
- ~`io/src/gpio.rs` contains the `GpioPin` trait and the hardware implementation using `embassy`. (This could be two separate files?)~ `io/src/gpio.rs` contains the `GpioPin` trait and the mock implementation.
- ~`io/src/mock_gpio.rs` contains the mock `GpioPin` implementation for use in testing~
- Each STM board that uses GPIO will have to implement it's own `GpioPin` in `boards/{board}/src/io`, e.g. `boards/stm32l476rg/src/io` as implemented on this PR.

`MockGpio` works by giving it a `heapless::Vec<bool, 10>` of test values. Every time `is_high()` is called it will move to the next value in the until they are all consumed and it will just stay on its last value. (This was done because we can only have one mutable reference in Rust so we can't just modify it since `Keyence` owns it.)

**To do:**

- [x] Settle on a code structure
- ~[ ] Test with Keyence sensor on an STM board~


boards/navigation/src/main.rs
```diff
@@ -0,0 +1,23 @@
...
+use {defmt_rtt as _, panic_probe as _};
+
+#[embassy_executor::main]
+async fn main(_spawner: Spawner) -> ! {
+    let p = embassy_stm32::init(Default::default());
+    let gpio_input = Input::new(p.PA0, Pull::Up);
+    let gpio = EmbassyGpio::new(gpio_input);
+    let mut keyence = Keyence::new(gpio);
+
+    loop {
```
kshxtij: This isn't something we've clarified yet, but i think each sensor should have it's own task since the main task (which runs with the highest prio by default iirc) would be incharge of the can communication for that board (obviously this isn't something done yet but i think just implementing everything as it's own task is decent enough future proofing).

kshxtij: This can ofc always be a ToDoLater for as soon as we figure that portion of the architecture out, up to you.

kshxtij (commented):
Looks really good.

Idk if it would be worth adding more no_std vs std flags and just using std::Vec cause surely we are never using MockGpio in a no_std env but it's probably not worth the effort.

boards/navigation/src/main.rs
davidbeechey: Good point, will change :)

davidbeechey: Re no_std Vec stuff, I'm pretty sure we can't have std just for tests since they're co-located in the same file, but I could be wrong...

kshxtij: Ah you're probably right, i think cargo flags can let us do something with it by just declaring std as an external crate in the test mode but that's ugly and probably not worth it 👍

davidbeechey: Ah okay, we could do that if we really need std for something in the future

davidbeechey:
**Update on testing:** I have now tested it using the button (PC13) on the STM board to simulate a GPIO high and it works as expected.

H-Allen (approved):

licornes-fluos (approved):

davidbeechey:
Ok for some reason the user button PC13 on the STM32L476RG board is inverted, so it's high when it's not pressed and low when it is. I'm to assume this is just an issue with the STM32L476RG board and not a flaw in the code logic (?) because I think it worked on the F7 board before. ~I have inverted the `is_high()` output for now...~

davidbeechey:
Update: definitely works properly on the F7 board

PR #7, opened 2024-08-17
INFRA - Config

kshxtij:
Refactors the config implementation from 4793d46 on tel-server.

- Uses serde for traits based serial/deserial-isation. 
- Refactor to eliminate unions and use a combination of enums and structs instead to prevent the need for unsafe {} and hacky matches.
- Requires slight change in config nesting to allow for enum fields.
- Include tests for core functionality to ensure all enums and structs work are serialised correctly.

Screenshot 2024-08-17 at 21 33 17


davidbeechey (approved):
Looks great :)

TomLonergan03 (approved):

PR #9, opened 2024-08-20
INFRA - GitHub Actions CI

davidbeechey:
Add basic "Cargo Build & Test" and "Linter" workflows. 

**Linter:**
- Runs clippy
- Checks for TODOs
- Enforces LF

~(Currently fails because we have no Rust to build...)~

To Do:

- [x] Spell check


kshxtij:
LGTM, some other random rust CI tools i've run into which you could consider are:

1. https://crates.io/crates/typos
2. https://crates.io/crates/cargo-audit
3. https://rust-lang.github.io/rustfmt/ to enforce code style (the rust lsp respects a root rustfmt.toml as well and this can enforce line endings)

There's also yamlfmt if we forsee ourself using lots of yaml configs but otherwise it's probably not needed.

kshxtij (approved):

TomLonergan03 (approved):
lgtm

PR #10, opened 2024-08-20
TEL - Add Telemetry code from 2024 (HYPE-16)

davidbeechey:
Copies across the Telemetry code from `hyped-2024` from commit [2e2ba1b](https://github.com/Hyp-ed/hyped-2024/commit/2e2ba1bddf8b105dcbd984567e9d95a14b9eb90f) (before the "Cart Changes" PR was merged which broke a few things...)

~Currently uncertain whether we will be using the current NestJS server or migrating to a new Rust one. We can always merge as is now (including the current server implementation) and remove it later on if we decide to go with Rust.~ Decision has been made to keep the TypeScript server.

This is probably a good place to pause and let @arjunnaha have a look over the code and make any minor changes he wants to make before we merge this (although anything substantial should probably be done in a new PR).


arjunnaha (approved):
Think we just merge as is, and then we can deal with the fallout later

PR #11, opened 2024-08-21
INFRA - CAN Board Testing Code (HYPE-15)

kshxtij:
This PR implements a CAN stm32 test bench. Currently configures CAN3 on the board to send messages that include a constant 0xdeadbeef and then counting up.

⚠️ This PR introduces a commit pin on the embassy libraries as the Cargo mirror for embassy-stm32 hasn't been updated for 7 months, if this isn't something we want to do I can remove the pin and reimplement this in the old library standard.

TODO: 
- [x] Test this basic can implementation
- [ ] Come up with a more robust interface for everything required with can probably under IO/ after the keyence PR is merged
- [ ] Write the sending and receiving tasks generically as this will be required on every board used.
- [ ] Can Messages? but that can be outside of the scope of this PR depending on complexity
- [ ] Can Logging?    ^^ likewise


boards/can_tester/Cargo.toml
davidbeechey, line 1: should probably put this `git` pin on the Wiki somewhere

davidbeechey (commented):
Looking good, will message you and Samuel (and sensors PLs too probs) to arrange another lab test with CAN0 to see if that works

kshxtij:
This was the latest code from the lab session we had a couple of weeks ago now, a lot of commented code but the remaining code in the board folder was the minimal viable example that we got working with a BBB connected.

davidbeechey:
**Tested and working (07/11/24):** CAN sending from one board and receiving on another. 

**Still to do:**
- [ ] Work out why receiver only receives 2 messages and then stops (probably need to clear a buffer or something somewhere)
- [ ] Work out why only CAN1 works, not CAN2 and CAN3 (we're planning to have 2 different CAN buses so this is important). Current theory is there's a problem with the Rx interrupt.
- [ ] Create nice abstraction for CAN

davidbeechey (approved):

davidbeechey:
CAN messaging is being broken down into smaller PRs. This one is now just the basic testing code for CAN on the STM boards. PRs to come are:
- CAN message format/structure
- CAN sending task
- CAN receiving task
- CAN to MQTT

TomLonergan03 (approved):
looks good

PR #12, opened 2024-08-21
SNS - I2C Temperature (HYPE-14)

davidbeechey:
I2C temperature sensor (STTS22H) implementation in Rust for STM32. The specific board it's currently implemented for is the STM32L476RG (for ease of testing), but other boards can be very easily supported.

Data sheet: [https://www.st.com/resource/en/datasheet/stts22h.pdf](https://www.st.com/resource/en/datasheet/stts22h.pdf)

**Changes:**

- New `Temperature` struct which handles the logic for the STTS22H sensor given a `HypedI2c` struct. 
- `HypedI2c` is an abstraction of the Embassy I2C lib to make it agnostic to board type and to support adding a mock I2C implementation if needed in the future.

**New boards added:**
- The `boards/` directory contains all of the STM32 boards we will be supporting. Currently this is just `boards/stm32l476rg`.
- Boards have three main functions: implementing IO for the specific board in the `src/io` folder, defining tasks that can be run in `src/tasks` (e.g. here we have a `read_temp` task which can be spawned which handles all of the setup and loop logic), and `src/bin` which includes the executables which can be flashed onto the boards themselves.

_Tested and working!_


lib/sensors/src/temperature.rs
```diff
@@ -0,0 +1,122 @@
...
+        };
+    }
+}
+
+/// Represents the possible I2C addresses for the STTS22H temperature sensor
+#[repr(u8)]
+enum TemperatureAddresses {
+    Address3f = 0x3f,
+    // Other possible addresses
+    #[allow(dead_code)]
```
kshxtij: Just allow it for the whole enum at this point, instead of per arm imo. 

```rust
#[allow(dead_code)]
enum TemperatureAddresses {
...
}
```

lib/sensors/src/temperature.rs
```diff
@@ -0,0 +1,122 @@
...
+const STTS22H_DATA_TEMP_H: u8 = 0x07;
+const STTS22H_STATUS: u8 = 0x05;
+
+// Values to check the status of the temperature sensor from the STTS22H_STATUS register
+const STTS22H_STATUS_BUSY: u8 = 0x01;
+const STTS22H_TEMP_OVER_UPPER_LIMIT: u8 = 0x02;
+const STTS22H_TEMP_UNDER_LOWER_LIMIT: u8 = 0x04;
+
+// These settings set the sensor to continuous mode, sets IF_ADD_INC, and sets the sampling rate to 200Hz
+const STTS22H_CONFIG_SETTINGS: u8 = 0x3c;
```
kshxtij, line 112: Just a sanity check, this one value configures all of that?

lib/sensors/src/temperature.rs
```diff
@@ -0,0 +1,122 @@
...
+                }
+            };
+        let temperature_low_byte =
+            match self.i2c.read_byte(self.device_address, STTS22H_DATA_TEMP_L) {
+                Some(byte) => byte,
+                None => {
+                    return None;
+                }
+            };
+        let combined = ((temperature_high_byte as u16) << 8 | temperature_low_byte as u16) as f32;
```
kshxtij:
```rust
let combined: f32 = ....
``` 
is better than the as at the end for readability imo but up to you

lib/core/src/types.rs
```diff
@@ -13,3 +13,20 @@ impl DigitalSignal {
...
+    use super::*;
+
+    #[test]
+    fn test_from_bool() {
+        assert_eq!(DigitalSignal::from_bool(true), DigitalSignal::High);
+        assert_eq!(DigitalSignal::from_bool(false), DigitalSignal::Low);
+    }
+}
+
+#[derive(Debug)]
```
kshxtij: Normally you group all the tests at the end and just declarations and stuff at the top, maybe we wanna do that?

kshxtij (changes requested):

lib/sensors/src/temperature.rs
```diff
@@ -0,0 +1,122 @@
...
+///
+/// Data sheet: https://www.st.com/resource/en/datasheet/stts22h.pdf
+pub struct Temperature<T: HypedI2c> {
+    i2c: T,
+    device_address: u8,
+}
+
+impl<T: HypedI2c> Temperature<T> {
+    /// Create a new instance of the temperature sensor and attempt to configure it
+    pub fn new(mut i2c: T) -> Result<Self, ()> {
```
kshxtij: probably take in a `TemperatureAddresses` value as an argument cause this should be generic over all 4 addresses

kshxtij (changes requested):

lib/sensors/src/temperature.rs
davidbeechey: Honestly not sure

lib/sensors/src/temperature.rs
davidbeechey: This doesn't work without the `as` at then end 🤔

lib/sensors/src/temperature.rs
davidbeechey: Done :)

licornes-fluos (approved):

lib/sensors/src/temperature.rs
davidbeechey: I think this is correct based on page 15 of the datasheet

kshxtij (approved):
LGTM, hopefully won't need to keep i2c_temp_test once the debugger is done

H-Allen (approved):

PR #25, opened 2024-10-10
SNS - Initial ADC Implementation

H00pyFr00d:
Converted the `adc.hpp` from [2024](https://github.com/Hyp-ed/hyped-2024/blob/master/lib/io/adc.hpp). David and I discussed and realised that no ADC exclusive sensor was ever used last year, and so only the MUX code has ever been used

As such this is being temporarily abandoned, the work done thusfar (however minimal) is being kept here

(Update)

We determined that it was worth implementing, though because Embassy has ADC functionality most of the C++ implementation could be ignored

I've implemented something basic for ADC, but it's not been tested and will need looked at by somebody else


lib/io/src/adc.rs
davidbeechey, line 1: We can't import `AnyAdcChannel` here because we don't have `embassy_stm32` in `lib/io`, so I think you'll have to implement the channel differently

lib/io/stm32l476rg/src/adc.rs
davidbeechey, line 1: According to the [documentation](https://docs.embassy.dev/embassy-stm32/git/stm32g031c8/adc/struct.Adc.html#method.blocking_read) for `blocking_read` it returns a `u16`

davidbeechey (changes requested):
A few changes are needed now because of the change in repo structure. You'll need to move the `lib/io/stm32l476rg/src/adc.rs` into `boards/stm32l476rg/src/io`. (You need to open the `boards/stm32l476rg` folder in VS Code to get IDE features from Rust Analyzer.) Not sure about the channel...

H00pyFr00d:
ADC implementation which passes `cargo build`, should (in theory) work

boards/stm32l476rg/src/io/adc.rs
```diff
@@ -0,0 +1,22 @@
+use embassy_stm32::{adc::Adc, adc::AnyAdcChannel, adc::Instance};
+use hyped_io::adc::HypedAdc;
+
+pub struct Stm32l476rgAdc<'d, T: Instance> {
+    adc: Adc<'d, T>,
+    pin: AnyAdcChannel<T>,
```
davidbeechey: rename `pin` to `channel` everywhere here

boards/stm32l476rg/src/io/adc.rs
```diff
@@ -0,0 +1,22 @@
...
+}
+
+impl<'d, T: Instance> HypedAdc for Stm32l476rgAdc<'d, T> {
+    /// Reading from ADC requires the peripheral pin to read from
```
davidbeechey:
```suggestion
    /// Read a value from the ADC channel
```

lib/io/src/adc.rs
```diff
@@ -0,0 +1,4 @@
+// ADC trait used to abstract the ADC peripheral
```
davidbeechey:
```suggestion
/// ADC trait used to abstract the ADC peripheral
```
Just need a triple `/` for Rust doc

davidbeechey (changes requested):
Figured out the channel thing:
- `embassy_stm32::adc::Adc::new` takes in a peripheral which is the ADC on the board (there is more than one), e.g. `p.ADC1`
- Then the `AnyAdcChannel` input to `Adc::blocking_read` is a pin which implements the `AnyAdcChannel` trait for selected ADC which was passed into `new`.

Basically makes sense to rename `pin` to `channel` anyway but otherwise looks perfect. To instantiate it would be this (I've just checked that this compiles properly):
```rust
let adc = Adc::new(p.ADC1);
let hyped_adc = Stm32l476rgAdc::new(adc, p.PA0.degrade_adc());
```

Also fix typos + formatting, then mark as ready for review :)

davidbeechey (approved):

TomLonergan03 (approved):

PR #26, opened 2024-10-14
SNS - I2C ToF

Aux1r:
initial commit for time of flight sensor code, hope the folder structure is okay


Aux1r:
Datasheet available here for the VL6180V1 ToF Sensor: https://www.st.com/en/imaging-and-photonics-solutions/vl6180.html#overview

Aux1r:
- Added functions for i2c trait to allow reading and writing to registers with 16 bit addresses.
- Created new() function, start_ss_measure() function which starts a singleshot measurement, poll_range() function that polls to see if a new range reslt is available, read_range() that reads this value.

A lot of these functions were implemented using the 'example code' provided in the VL6180 application sheet: https://www.st.com/resource/en/application_note/an4545-vl6180x-basic-ranging-application-note-stmicroelectronics.pdf

**TODO:**

- Implement continuous mode measurements
- Verify decoding of RESULT_RANGE_VAL is appropriate (and actually returns us the value in mm)
- Verify poll_range() function works (not sure how Rust likes the while loop)
- Add clear_interrupts() function and checks for if the system is 'fresh out of reset' (refer to Application Sheet)

lib/io/src/i2c.rs
```diff
@@ -21,4 +21,11 @@ pub trait HypedI2c {
         register_address: u8,
         data: u8,
     ) -> Result<(), I2cError>;
+    fn read_byte_16(&mut self, device_address: u8, register_address: u16) -> Option<u8>;
```
davidbeechey: The C++ code only required 16-bit register addresses for the `writeByteToRegister` function, is it needed for `read_byte` too?

lib/sensors/src/tof.rs
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    pub fn poll_range(&mut self) {
+        let status_byte = 
+        match self.i2c.read_byte(self.device_address, RESULT_INTERR_STATUS_GPIO) {
+            Some(byte) => byte,
+            None => 0, // not sure about returning 0 for None - will this somehow break stuff?
+        };
+        let range_status = status_byte & 0x07;
+        while (range_status != 0x04) {
```
davidbeechey: Yeah I'm not sure about this loop, there must be an alternative way of doing it which is less scary

davidbeechey (changes requested):

lib/io/src/i2c.rs
Aux1r: good point; `read_byte_16` isn't needed and will be removed

boards/stm32l476rg/src/tasks/tof.rs
```diff
@@ -0,0 +1,34 @@
...
+    let mut tof_sensor = TimeOfFlight::new(&mut hyped_i2c, ToFAddresses::Address29).expect(
+        "Failed to create Time of Flight sensor. Check the wiring and I2C address of the sensor.",
+    );
+
+    loop {
+        defmt::info!("Starting single shot measurement");
+        tof_sensor.start_ss_measure();
+        defmt::info!("Polling for range");
+        tof_sensor.poll_range().await;
+        defmt::info!("Reading range");
```
davidbeechey: I think a nicer API for this would be to just call `start_ss_measure` and that also handle the polling internally? Or go even further and just have `read` which does everything

lib/sensors/src/tof.rs
```diff
@@ -0,0 +1,386 @@
...
+        // Note: the sensor will either need to be power cycled or the SYS_FRESH_OUT_RESET register will need to be written to 0x01 if it has already been configured
+        let mut boot_status = i2c
+            .read_byte_16(device_address, SYSTEM_FRESH_OUT_OF_RESET)
+            .unwrap_or_default();
+        while boot_status != 1 {
+            boot_status = i2c
+                .read_byte_16(device_address, SYSTEM_FRESH_OUT_OF_RESET)
+                .unwrap_or_default();
+        }
+        defmt::info!("ToF sensor booted");
```
davidbeechey: Thinking about this again: To take advantage of the `SYSTEM_FRESH_OUT_OF_RESET` register, you could just check if it is configured and if it is, skip initialising all of the registers (just extract that logic into an `init_sensor` (or similar) function

lib/sensors/src/tof.rs
```diff
@@ -0,0 +1,386 @@
...
+    pub fn start_ss_measure(&mut self) -> Result<i32, ToFError> {
+        // start single shot measurement
+        if let Err(e) = self.i2c.write_byte_to_register_16(
+            self.device_address,
+            SYSRANGE_START,
+            SYSRANGE_START_SS_VAL,
+        ) {
+            return Err(ToFError::I2cError(e));
+        }
+        Ok(1)
```
davidbeechey: You can just return unit (`()`) here instead

lib/sensors/src/tof.rs
```diff
@@ -0,0 +1,386 @@
...
+            .unwrap_or(0);
+        let mut range_status = status_byte & 0x07;
+        while range_status != 0x04 {
+            defmt::debug!("Range status: {:?}", range_status);
+            range_status = self
+                .i2c
+                .read_byte_16(self.device_address, RESULT_INTERR_STATUS_GPIO)
+                .unwrap_or_default()
+                & 0x07;
+            Timer::after(Duration::from_millis(100)).await;
```
davidbeechey: Feel free to remove this and make the function synchronous again

lib/sensors/src/tof.rs
```diff
@@ -0,0 +1,386 @@
...
+        Ok(1)
+    }
+
+    pub fn check_reset(&mut self) -> bool {
+        let reset_value = self
+            .i2c
+            .read_byte_16(self.device_address, SYS_FRESH_OUT_RESET)
+            .unwrap_or(0);
+        reset_value == 1
+    }
```
davidbeechey: Didn't notice this earlier, you can either remove or use this in the `new` function instead

lib/sensors/src/tof.rs
```diff
@@ -0,0 +1,386 @@
...
+        // at the end clear interrupts
+        if let Err(e) = self.i2c.write_byte_to_register_16(
+            self.device_address,
+            SYS_INTERRUPT_CLEAR,
+            CLEAR_INTERRUPTS_VAL,
+        ) {
+            return Err(ToFError::I2cError(e));
+        }
+        Ok(1)
+    }
```
davidbeechey: When will this be called?

lib/sensors/src/tof.rs
```diff
@@ -0,0 +1,386 @@
...
+    pub fn clear_interrupts(&mut self) -> Result<i32, ToFError> {
+        // at the end clear interrupts
+        if let Err(e) = self.i2c.write_byte_to_register_16(
+            self.device_address,
+            SYS_INTERRUPT_CLEAR,
+            CLEAR_INTERRUPTS_VAL,
+        ) {
+            return Err(ToFError::I2cError(e));
+        }
+        Ok(1)
```
davidbeechey: Again, you can return unit instead

lib/sensors/src/tof.rs
```diff
@@ -0,0 +1,386 @@
...
+// can't find reference to 0x003f address for light and dark gain in datasheet from the application note
+const SYSRANGE_VHV_REPEAT_RATE: u16 = 0x031;
+// can't find reference to 0x0041, see above.
+const SYSRANGE_VHV_RECALIBRATE: u16 = 0x002e;
+const SYSRANGE_INTERMEASURE_PERIOD: u16 = 0x01b;
+// same story with 0x003e
+const SYS_INTERRUPT_CONFIG_GPIO: u16 = 0x014;
+const SYSRANGE_START: u16 = 0x018;
+const RESULT_INTERR_STATUS_GPIO: u16 = 0x04f;
+const SYS_FRESH_OUT_RESET: u16 = 0x016;
```
davidbeechey: This is a duplicate now, my bad

davidbeechey (changes requested):
This is very nearly done now! The tests aren't working properly (not sure why). A few comments here to fix and think about the API.  Apart from those functionality points, would be good to add Rust doc comments to all of your functions

boards/stm32l476rg/src/tasks/tof.rs
```diff
@@ -0,0 +1,34 @@
...
+
+        match tof_sensor.read_range() {
+            Some(range) => {
+                defmt::info!("Range: {:?}", range)
```
davidbeechey: Forgot to say: add units here?

lib/sensors/src/tof.rs
Aux1r: Clear Interrupts is called at the end of the reading cycle as seen on page 22 of the [application sheet](https://www.st.com/resource/en/application_note/an4545-vl6180x-basic-ranging-application-note-stmicroelectronics.pdf); I'll add it to the ToF task since it's missing from there

boards/stm32l476rg/src/tasks/tof.rs
Aux1r: Poll_range functionality will be added to start_ss_measure() so we don't have to call these separately to handle starting the measurement and polling.

lib/io/src/i2c.rs
```diff
@@ -38,18 +45,42 @@ pub mod mock_i2c {
...
+        /// Reads a byte by looking up the device address and register address in the map
+        fn read_byte_16(&mut self, device_address: u8, register_address: u16) -> Option<u8> {
             self.values
                 .get(&(device_address, register_address))
                 .copied()
                 .unwrap()
         }
 
-        /// Writes a byte to the map so that it can be read later to check the value
+        /// Writes a byte to the map 9with 8-bit register address) so that it can be read later to check the value
```
TomLonergan03: typo: 9 should be (

lib/sensors/src/tof.rs
```diff
@@ -0,0 +1,414 @@
...
+        defmt::info!("ToF sensor configured");
+
+        Ok(Self {
+            i2c,
+            device_address,
+        })
+    }
+
+    pub fn start_ss_measure(&mut self) -> Result<(), ToFError> {
+        // start single shot measurement
```
TomLonergan03: rename function to `start_single_shot_measurement`, this comment isn't necessary then

lib/sensors/src/tof.rs
```diff
@@ -0,0 +1,414 @@
...
+            Some(byte) => byte,
+            None => {
+                return None;
+            }
+        };
+        Some(range_byte)
+    }
+
+    pub fn start_cts_measure(&mut self) -> Result<(), ToFError> {
+        // start continuous measurement
```
TomLonergan03: same here as with single shot

lib/sensors/src/tof.rs
```diff
@@ -0,0 +1,414 @@
...
+    }
+
+    pub fn read_range(&mut self) -> Option<u8> {
+        let range_byte = match self.i2c.read_byte_16(self.device_address, RESULT_RANGE_VAL) {
+            Some(byte) => byte,
+            None => {
+                return None;
+            }
+        };
+        Some(range_byte)
```
TomLonergan03: this is unnecessary, you can just do 
```rs
self.i2c.read_byte_16(self.device_address, RESULT_RANGE_VAL)
```
rn you unwrap the option to immediately wrap it back up

lib/sensors/src/tof.rs
```diff
@@ -0,0 +1,414 @@
...
+
+    pub fn check_reset(&mut self) -> bool {
+        let reset_value = self
+            .i2c
+            .read_byte_16(self.device_address, SYS_FRESH_OUT_RESET)
+            .unwrap_or(0);
+        reset_value == 1
+    }
+
+    pub fn clear_interrupts(&mut self) -> Result<(), ToFError> {
```
TomLonergan03: a bunch of these `pub fn`s aren't actually used anywhere. will they be used in the actual task outside of the demo ones? or are they just here for completeness?

lib/sensors/src/tof.rs
```diff
@@ -0,0 +1,414 @@
...
+        let mut i2c_values = FnvIndexMap::new();
+        let _ = i2c_values.insert(
+            (ToFAddresses::Address29 as u8, SYS_FRESH_OUT_RESET),
+            Some(1),
+        );
+        let _ = i2c_values.insert((ToFAddresses::Address29 as u8, RESULT_RANGE_VAL), Some(255));
+        let mut i2c = MockI2c::new(i2c_values);
+        let mut tof = TimeOfFlight::new(&mut i2c, ToFAddresses::Address29).unwrap();
+        assert_eq!(tof.read_range(), Some(255));
+    }
```
TomLonergan03: are these 3 tests hitting different paths/behaviours? they look like they are just duplicates of each other

lib/sensors/src/tof.rs
```diff
@@ -0,0 +1,414 @@
...
+        // Check that the sensor has powered up
+        // Note: the sensor will either need to be power cycled or the SYS_FRESH_OUT_RESET register will need to be written to 0x01 if it has already been configured
+        let mut boot_status = i2c
+            .read_byte_16(device_address, SYSTEM_FRESH_OUT_OF_RESET)
+            .unwrap_or_default();
+        while boot_status != 1 {
+            boot_status = i2c
+                .read_byte_16(device_address, SYSTEM_FRESH_OUT_OF_RESET)
+                .unwrap_or_default();
+        }
```
TomLonergan03: i'm not keen on these unbounded initalisation loops, but i'm not sure what the policy is on them across the repo. i like the nasa 10 rules thing of having a fixed upper bound on all loops, so having a maximum number of attempts before you give up. @davidbeechey opinions?

lib/sensors/src/tof.rs
```diff
@@ -0,0 +1,414 @@
...
+            device_address,
+            SYSRANGE_INTERMEASURE_PERIOD,
+            SYSRANGE_INTERMEASURE_PERIOD_VAL,
+        ) {
+            return Err(ToFError::I2cError(e));
+        }
+
+        // if let Err(e) = i2c.write_byte_to_register_16(device_address, 0x003e, 0x31) {
+        //     return Err(ToFError::I2cError(e));
+        // }
```
TomLonergan03: several bits of commented out code in this function, it should be uncommented or removed

lib/sensors/src/tof.rs
```diff
@@ -0,0 +1,414 @@
...
+        let range_byte = match self.i2c.read_byte_16(self.device_address, RESULT_RANGE_VAL) {
+            Some(byte) => byte,
+            None => {
+                return None;
+            }
+        };
+        Some(range_byte)
+    }
+
+    pub fn start_cts_measure(&mut self) -> Result<(), ToFError> {
```
TomLonergan03: it would be good to have doc comments (`/// docs here`) for this and `start_ss_measure` explaining why you want one instead of the other

TomLonergan03 (changes requested):
looks good. i didn't read the datasheet so haven't commented on correctness stuff there

2 things:
- in general i think you should rename the "tof"s in file names and code to be "time_of_flight", just for sake of clarity.
- a lot of the `TimeOfFlight` impl methods are kinda unclear on what they do, so there's probably better names for them or if not add doc comments

lib/sensors/src/tof.rs
davidbeechey: Agreed, I'd say we're confident after max. 10 seconds that something is wrong so can `panic!`

lib/sensors/src/tof.rs
Aux1r: Extracted register config loading logic into a standalone function and am now calling it inside of an if-statement that checks whether the system is freshly reset or not.

lib/sensors/src/tof.rs
Aux1r: In the newest version of the PR, there's only 2 tests. test_range_read_0 and test_range_read_255 are the same as each other but both test the upper and lower bounds of what the RESULT_RANGE_VAL can possibly return. If there's no need to test different values, happy to just cut this down to one test that tests if it can read a value like 255.

lib/sensors/src/tof.rs
Aux1r: Removed some redundant `pub fns` and added doc comments to the others to clarify what they do.

boards/stm32l476rg/src/tasks/time_of_flight.rs
```diff
@@ -0,0 +1,39 @@
...
+    )
+    .expect(
+        "Failed to create Time of Flight sensor. Check the wiring and I2C address of the sensor.",
+    );
+
+    loop {
+        // defmt::info!("Starting single shot measurement");
+        // defmt::info!("Polling for range");
+        time_of_flight_sensor.start_single_shot_measurement();
+        // defmt::info!("Started single shot measurement");
```
TomLonergan03: commented code

lib/sensors/src/time_of_flight.rs
```diff
@@ -0,0 +1,362 @@
...
+                .unwrap_or_default()
+                & 0x07;
+        }
+        Ok(())
+    }
+
+    /// Read range from the RESULT_RANGE_VAL register and return it
+    pub fn read_range(&mut self) -> Option<u8> {
+        self.i2c.read_byte_16(self.device_address, RESULT_RANGE_VAL)
+    }
```
TomLonergan03: is there a reason to have these as separate functions? if you always read the range straight after starting the measurement it might as well be one public function, to keep the public api cleaner

lib/sensors/src/time_of_flight.rs
```diff
@@ -0,0 +1,362 @@
...
+            CLEAR_INTERRUPTS_VAL,
+        ) {
+            return Err(TimeOfFlightError::I2cError(e));
+        }
+        Ok(())
+    }
+}
+
+/// Standalone function used in the 'new' method of the TimeOfFlight sensor, loads all of the configs according to SR03 settings
+fn load_config<T: HypedI2c>(i2c: &mut T, device_address: u8) -> Result<(), TimeOfFlightError> {
```
TomLonergan03: @davidbeechey do you have repo rules on what is a private method of an `impl` vs what goes in the file? if there isn't then this is fine

TomLonergan03 (commented):
looks very good, couple of notes

lib/sensors/src/tof.rs
TomLonergan03: cos both are returning `u8` the 0-255 bound is already enforced (in debug profile at least) so you should be able to merge them.

davidbeechey: rust moment :crab:

lib/sensors/src/time_of_flight.rs
davidbeechey: Agreed, I think I mentioned this at some point but maybe in person and I forgot to add it here

lib/sensors/src/time_of_flight.rs
davidbeechey: I think this should be a private method on the struct and not in the file

davidbeechey (changes requested):

H-Allen (approved):

TomLonergan03 (approved):

davidbeechey (approved):

PR #27, opened 2024-10-14
SNS - LED Driver

siny-over-cosy:


lib/sensors/led_driver.rs
```diff
@@ -0,0 +1,23 @@
...
+impl<T: HypedI2c> LedDriver<T> {
+    /// Create new instance of LED driver and attempt to configure
+    pub fn initialise(mut i2c: T, device_address: ledConfigRegister) -> Result<Self, ledDriverError> {
+        
+    }
+}
+
+
+
+const defaultLedDriverAddress: u8 = 0x30;
```
davidbeechey: The convention for Rust variable names is `snake_case` for normal variables and `SHOUTING_CASE` for constants (like these)

lib/sensors/led_driver.rs
```diff
@@ -0,0 +1,23 @@
+use hyped_io::i2c::{HypedI2c, I2cError};
+
+pub struct LedDriver<T: HypedI2c> {
+    i2c: T,
+    device_address: u8,
+}
+
+impl<T: HypedI2c> LedDriver<T> {
+    /// Create new instance of LED driver and attempt to configure
+    pub fn initialise(mut i2c: T, device_address: ledConfigRegister) -> Result<Self, ledDriverError> {
```
davidbeechey:
```suggestion
    pub fn new(mut i2c: T, device_address: ledConfigRegister) -> Result<Self, ledDriverError> {
```

lib/sensors/led_driver.rs
davidbeechey, line 1: This file should be in `lib/sensors/src/led_driver.ts`, and you'll need to add `pub mod led_driver;` to `lib/sensors/src/lib.rs`.

davidbeechey (commented):
Good start! A few wee things to fix. Would be good to add a description + link to the datasheet in the PR description, Linear issue description and as a Rust doc comment (using `///`) for `LedDriver`

lib/core/Cargo.lock
davidbeechey, line 1: could you restore this by running `cargo build`?

lib/sensors/led_driver
davidbeechey, line 1: delete this

lib/sensors/src/led_driver.rs
```diff
@@ -0,0 +1,247 @@
+/// LED Driver
+
+/// When setting colours and brightness of LEDs, use the set_led_colours() function.
+///     LEDs needing to be set are controlled by the addresses in either LED_CONFIG0 or LED_CONFIG1 registers
+///     These addresses can be toggled with hexadecimal (E.G: registers 11101000 are set by the hex value E8)
+
+/// link to data sheet (https://www.ti.com/lit/ds/symlink/lp5036.pdf?HQS=dis-mous-null-mousermode-dsf-pf-null-wwe&ts=1698441544495&ref_url=https%253A%252F%252Fwww.mouser.co.uk%252F)
+///
```
davidbeechey: Good documentation, but move it to above `pub struct LedDriver...` below

davidbeechey (changes requested):
This is looking really good, it just needs a task implementation (like for the temperature sensor in `boards/stm32l476rg/src/tasks/temperature.rs`). We can have a look at this in tomorrow's meeting. Then it'll be ready for testing!

lib/sensors/src/led_driver.rs
siny-over-cosy: Done + added additional documentation.

PR #28, opened 2024-10-14
SNS - I2C Accelerometer

miksquared:
Translate last year's accelerometer code into Rust.
Data sheet: https://www.st.com/resource/en/datasheet/lis2ds12.pdf


davidbeechey (commented):
Correct structure! You'll need to add `pub mod accelerometer;` to `lib/sensors/src/lib.rs` to include it (this will activate IDE features too)

lib/sensors/src/accelerometer.rs
```diff
@@ -0,0 +1,67 @@
...
+pub struct Accelerometer<T: HypedI2c> {
+    i2c: T,
+    device_address: u8,
+}
+
+impl<T: HypedI2c> Accelerometer<T> {
+    /// Create a new instance of the accelerometer and attempt to configure it
+    pub fn new(mut i2c: T, device_address: AccelerometerAddresses) -> Result<Self, AccelerometerError> {
+        let device_address = device_address as u8;
+        let _ctrl_1_result = match i2c.write_byte_to_register(device_address, LIS2DS12_CTRL1_ADDRESS, LIS2DS12_CTRL1_VALUE) {
```
davidbeechey:
```suggestion
        match i2c.write_byte_to_register(device_address, LIS2DS12_CTRL1_ADDRESS, LIS2DS12_CTRL1_VALUE) {
```
The `match` is handling the `Result` type from `write_byte_to_register`, so you don't have to have this

davidbeechey (changes requested):
Looking good, thanks for including a detailed comment. Could you also add a brief description + data sheet link to the PR description and Linear issue?

lib/sensors/src/accelerometer.rs
miksquared: Removed assignments in commit 685f9e4

lib/sensors/src/accelerometer.rs
```diff
@@ -0,0 +1,153 @@
...
+    pub fn read(&mut self) -> Option<(f32, f32, f32)> {
+
+        // Read the low and high bytes of the acceleration and combine them to get the acceleration for each axis
+        let x_low_byte =
+            match self.i2c.read_byte(self.device_address, LIS2DS12_OUT_X_L) {
+                Some(byte) => byte,
+                None => {
+                    return None;
+                }
+            };
```
TomLonergan03: I believe this can be
```rust
let x_low_byte = self.i2c.read_byte(self.device_address, LIS2DS12_OUT_X_L)?;
```
as `?` will either give you the value in `Some(...)` or return early if it's `None`, which is more succinct

and the same for the rest of the reads

lib/sensors/src/accelerometer.rs
```diff
@@ -0,0 +1,153 @@
...
+}
+
+impl<T: HypedI2c> Accelerometer<T> {
+    /// Create a new instance of the accelerometer and attempt to configure it
+    pub fn new(mut i2c: T, device_address: AccelerometerAddresses) -> Result<Self, AccelerometerError> {
+        let device_address = device_address as u8;
+        match i2c.write_byte_to_register(device_address, LIS2DS12_CTRL1_ADDRESS, LIS2DS12_CTRL1_VALUE) {
+            Err(e) => return Err(AccelerometerError::I2cError(e)),
+            Ok(p) =>  p
+        };
```
TomLonergan03: You're not using `Ok(p) = p`  so you should replace it with `_ => ()` imo

or possibly even
```rust
        if let Err(e) =
            i2c.write_byte_to_register(device_address, LIS2DS12_CTRL1_ADDRESS, LIS2DS12_CTRL1_VALUE)
        {
            return Err(AccelerometerError::I2cError(e));
        };
```
but that feels less nice to me, maybe @davidbeechey has a take?

lib/sensors/src/accelerometer.rs
```diff
@@ -0,0 +1,153 @@
...
+
+        // Return Self only if all values are written successfully
+        Ok(Self{i2c, device_address})
+    }
+
+    /// Read the acceleration for each axis and return them as a floating point value in gs.
+    pub fn read(&mut self) -> Option<(f32, f32, f32)> {
```
TomLonergan03: is it preferred to return these as a tuple or have named fields in a struct to keep it clear which is x/y/z?

lib/sensors/src/accelerometer.rs
```diff
@@ -0,0 +1,153 @@
...
+    Unknown
+}
+
+impl Status {
+    pub fn from_byte(byte: u8) -> Self {
+        match byte {
+            LIS2DS12_DATA_NOT_READY => Self::DataNotReady,
+            _ => Self::Ok,
+        }
+    }
```
TomLonergan03, line 138: is this correct? the manual seems to say that only the `DRDY` bit matters for the data being ready, can there be other flags set in the status register that you don't care about for determining readiness?

TomLonergan03 (commented):
looks good! there's a couple things i'm unsure on + i'm not too confident on the coding standards yous are using with rust, so ignore what I say as you wish lol

oh also run a format over it

lib/sensors/src/accelerometer.rs
davidbeechey: Yeah I think a struct is better because it's more explicit

lib/sensors/src/accelerometer.rs
davidbeechey: I think I prefer just doing `_ => ()`

davidbeechey (changes requested):
Looks good, just fix @TomLonergan03's suggestions. A couple of tests would be good too, like the ones for the temperature sensor (you might need to merge main into your branch again to get `MockI2c`)

lib/sensors/src/accelerometer.rs
```diff
@@ -0,0 +1,123 @@
...
+    pub fn check_status(&mut self) -> Status {
+        match self.i2c.read_byte(self.device_address, LIS2DS12_STATUS) {
+            Some(byte) => Status::from_byte(byte),
+            None => Status::Unknown,
+        }
+    }
+
+}
+
+pub struct AccelValues {
```
davidbeechey:
```suggestion
pub struct AccelerationValues {
```
Sorry being picky

davidbeechey (changes requested):
Very nearly there. Just tests now I think

TomLonergan03 (approved):
looks good, see if there's a way to test `DRDY` status bit when your in the lab, and do what clippy/rustfmt want

davidbeechey (approved):
Looks good, ready for testing!

lib/sensors/src/accelerometer.rs
```diff
@@ -7,56 +7,75 @@ use hyped_io::i2c::{HypedI2c, I2cError};
...
-        match i2c.write_byte_to_register(device_address, LIS2DS12_FIFO_CTRL_ADDRESS, LIS2DS12_FIFO_CTRL_VALUE) {
-            Err(e) => return Err(AccelerometerError::I2cError(e)),
-            _ => ()
-        };
-
+        if let Err(e) =
+            i2c.write_byte_to_register(device_address, LIS2DS12_CTRL1_ADDRESS, LIS2DS12_CTRL1_VALUE)
+        {
+            return Err(AccelerometerError::I2cError(e));
+        }
```
TomLonergan03: maybe make a macro for this? idk if there's a nice way to generalise it across the different i2c users @davidbeechey?

davidbeechey: Macro for this is in PR #58 :) I have tested it works here on the branch `test-i2c_macros_with_accel` and seems to work okay

boards/stm32f767zi/src/tasks/accelerometer.rs
```diff
@@ -0,0 +1,65 @@
...
+    accelerometer::{AccelerationValues, Accelerometer, AccelerometerAddresses, Status},
+    SensorValueRange::*,
+};
+
+/// Test task that reads the acceleration from the sensor and prints it to the console.
+#[embassy_executor::task]
+pub async fn read_acceleration() -> ! {
+    let p = embassy_stm32::init(Default::default());
+    let i2c = I2c::new_blocking(p.I2C1, p.PB8, p.PB9, Hertz(200_000), Default::default());
+    let mut hyped_i2c = Stm32f767ziI2c::new(i2c);
```
davidbeechey: The way we're creating I2C instances has changed a bit - have a look at `temperature.rs` at how it is done now.

Looks like:

```rust
let i2c = I2c::new_blocking(p.I2C1, p.PB8, p.PB9, Hertz(100_000), Default::default());
static I2C_BUS: StaticCell<I2c1Bus> = StaticCell::new();
let i2c_bus = I2C_BUS.init(Mutex::new(RefCell::new(i2c)));
let mut hyped_i2c = Stm32f767ziI2c::new(i2c_bus);
```

where you have the type def:

```rust
type I2c1Bus = Mutex<NoopRawMutex, RefCell<I2c<'static, Blocking>>>;
```

above the function definition

davidbeechey (changes requested):

davidbeechey (approved):
Looks good!

boards/stm32f767zi/src/tasks/accelerometer.rs
davidbeechey, line 1: rename to `read_accelerometer.rs`

boards/stm32f767zi/src/bin/mux_test.rs
```diff
@@ -0,0 +1,132 @@
...
+use static_cell::StaticCell;
+use {defmt_rtt as _, panic_probe as _};
+
+const MUX_ADDRESS: u8 = 0x70;
+
+type I2c1Bus = Mutex<NoopRawMutex, RefCell<I2c<'static, Blocking>>>;
+
+/// Task that reads the accelerometer from a specific channel of an I2C Mux.
+#[embassy_executor::task(pool_size = 4)]
+async fn read_accelerometer_from_mux(
```
davidbeechey: could you move this task into it's own file in `../tasks` called `read_accelerometer_from_mux` pls

davidbeechey (changes requested):
Just a few file changes.

Also could you modify this so that it uses the new `Watch` stuff?

boards/stm32f767zi/src/tasks/read_accelerometer.rs
```diff
@@ -0,0 +1,71 @@
...
+type I2c1Bus = Mutex<NoopRawMutex, RefCell<I2c<'static, Blocking>>>;
+
+/// Test task that reads the acceleration from the sensor and prints it to the console.
+#[embassy_executor::task]
+pub async fn read_acceleration() -> ! {
+    let p = embassy_stm32::init(Default::default());
+    let i2c = I2c::new_blocking(p.I2C1, p.PB8, p.PB9, Hertz(200_000), Default::default());
+    static I2C_BUS: StaticCell<I2c1Bus> = StaticCell::new();
+    let i2c_bus = I2C_BUS.init(Mutex::new(RefCell::new(i2c)));
+    let mut hyped_i2c = Stm32f767ziI2c::new(i2c_bus);
```
davidbeechey: One last thing: these should be passed as arguments to the task, rather than initialised here. Then it would be good to create board code in `/bin` that does all the initialisation and spawns the task

davidbeechey (changes requested):

boards/stm32f767zi/src/bin/accelerometer_test.rs
```diff
@@ -0,0 +1,85 @@
...
+                            accel_values.y,
+                            accel_values.z
+                        );
+                    }
+                },
+                None => {
+                    defmt::info!("Failed to read acceleration values.")
+                }
+            }
+            Timer::after(Duration::from_millis(100)).await;
```
TomLonergan03: is it correct that this is inside the `if let`?

boards/stm32f767zi/src/tasks/read_accelerometer.rs
```diff
@@ -0,0 +1,53 @@
...
+            Status::DataNotReady => {
+                defmt::warn!("Accelerometer is not ready to provide data");
+            }
+            Status::Unknown => {
+                panic!("Could not get the status of the accelerometer.");
+            }
+            Status::Ok => {}
+        }
+
+        sender.send(accelerometer.read());
```
TomLonergan03: should the read still happen if the accelerometer is not ready for the read?

boards/stm32f767zi/src/tasks/read_accelerometer_from_mux.rs
```diff
@@ -0,0 +1,132 @@
...
+                        accel_values.z
+                    );
+                }
+            },
+            None => {
+                defmt::info!("Failed to read acceleration values.");
+            }
+        }
+
+        Timer::after(Duration::from_secs(1)).await;
```
TomLonergan03: why is this a 1s sleep when the others are 100ms sleeps/frequencies? or are these all just needing to be "good enough" numbers for testing?

davidbeechey: I see this has just been updated, but also from now on the sampling frequencies will be defined from the `config/sensors.yaml` file

boards/stm32f767zi/src/tasks/read_accelerometer.rs
davidbeechey: @miksquared this is a good point, perhaps move `sender.send(accelerometer.read())` to inside the `Status::Ok => {}` block?

boards/stm32f767zi/src/tasks/read_accelerometer_from_mux.rs
```diff
@@ -0,0 +1,132 @@
...
+    }
+}
+
+#[embassy_executor::main]
+async fn main(spawner: Spawner) -> ! {
```
davidbeechey: this `main` function should be moved to a new `bin`: `bin/accelerometer_mux_test` (or similar)

boards/stm32f767zi/src/bin/accelerometer_test.rs
davidbeechey, line 1: I'm on a bit of an anti-abbreviation mission this year sorry, there's a few `accel`s about, could you change to full `accelerometer` pls?

davidbeechey (changes requested):

davidbeechey (approved):

H-Allen (approved):

PR #30, opened 2024-10-14
LOC - Preprocess Keyence

lylebuist:
Translate the preprocess_keyance file into Rust.


lib/localisation/src/lib.rs
```diff
@@ -1 +1 @@
-#![no_std]
```
misha7b, line 1: We aren't allowed to remove the no_std or things will break

lylebuist: Ah okay, can we import it to use print?

lib/localisation/src/preprocessing/keyence.rs
```diff
@@ -1 +1,48 @@
+use heapless::Vec;
 
+#[derive(PartialEq, Debug)]
+pub enum SensorChecks {
+    Acceptable,
+    Unnaceptable,
+}
+
+pub struct KeyenceAgrees {
```
davidbeechey, line 11: Quick Rust doc `///` comment here to explain what this does would be good

davidbeechey (changes requested):
Looks like the C++ code was checking whether the Keyence sensors deviated two times in a row, rather than just once? Also with this structure we'd need to create a new instance each time we check whether Keyence agrees. Instead the new Keyence values should be passed into `check_keyence_agrees`, but the struct `KeyenceAgrees` will now need to keep track of the previous agreement status (I think).

Is this correct @misha7b?

lib/localisation/src/preprocessing/keyence.rs
lylebuist: So would we want it to store the previous sensor check and compare with that too? That then gets updated at the end of each run with the new one? To handle the 2 in a row.

davidbeechey: Yeah exactly, first time the two new keyence values disagree you'd set `self.last_agreement_status` to `false` and if they disagree a second time then you'd trigger the failure (and it would reset to `true` when they agree)

lib/localisation/src/preprocessing/keyence.rs
```diff
@@ -1 +1,54 @@
...
+    keyence_data: Vec<bool, 2>,
```
davidbeechey:
```suggestion
```

lib/localisation/src/preprocessing/keyence.rs
```diff
@@ -1 +1,54 @@
...
+
+impl KeyenceAgrees {
+    pub fn new(keyence_data: Vec<bool, 2>) -> Self {
+        KeyenceAgrees {
+            keyence_data,
+            previous_keyance_agreement: true,
+        }
+    }
+
+    pub fn check_keyence_agrees(&mut self) -> SensorChecks {
```
davidbeechey:
```suggestion
    pub fn check_keyence_agrees(&mut self, keyence_data: Vec<bool, 2>) -> SensorChecks {
```
The new keyence data should be passed into this function rather than `new` because once a new `KeyenceAgrees` has been created we'll just keep calling `check_keyence_agrees` with new values

lib/localisation/Cargo.toml
```diff
@@ -5,3 +5,4 @@ edition = "2021"
 
 [dependencies]
 nalgebra = "0.33.0"
```
davidbeechey:
```suggestion
nalgebra = { version = "0.33.0", default-features = false }
```

davidbeechey (changes requested):

lib/localisation/src/preprocessing/keyence.rs
lylebuist: Clippy is suggesting that i change the impl to a Default one, however when I do this the impl no longer likes having functions.

davidbeechey: It just wants you to add:

```rust
impl Default for KeyenceAgrees {
    fn default() -> Self {
        Self::new()
    }
}
```

misha7b (approved):
Looks good to me!

lib/localisation/src/preprocessing/keyence.rs
```diff
@@ -1 +1,96 @@
...
+
+impl KeyenceAgrees {
+    pub fn new() -> Self {
+        KeyenceAgrees {
+            previous_keyance_agreement: true,
+        }
+    }
+
+    pub fn check_keyence_agrees(&mut self, keyence_data: Vec<bool, 2>) -> SensorChecks {
+        if keyence_data[0] != keyence_data[1] && !self.previous_keyance_agreement {
```
misha7b: I guess keyence_data[0] != keyence_data[1] doesnt have to be repeated twice but i dont think that will make any difference lol

lib/localisation/src/preprocessing/keyence.rs
```diff
@@ -1 +1,96 @@
+use heapless::Vec;
 
+#[derive(PartialEq, Debug)]
+pub enum SensorChecks {
+    Acceptable,
+    Unnaceptable,
+}
+
+/// This struct checks if both the current and previous keyance data is in disagreement
```
davidbeechey:
```suggestion
/// This struct checks if both the current and previous Keyence data is in disagreement
```

lib/localisation/src/preprocessing/keyence.rs
```diff
@@ -1 +1,96 @@
...
+impl Default for KeyenceAgrees {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl KeyenceAgrees {
+    pub fn new() -> Self {
+        KeyenceAgrees {
+            previous_keyance_agreement: true,
```
davidbeechey:
```suggestion
            previous_keyence_agreement: true,
```

davidbeechey (changes requested):
Sorry just spotted a couple of spelling things, then ready to merge!

davidbeechey (approved):

PR #32, opened 2024-10-14
SNS - Optical Flow

prakhars0206:


optical_flow.rs
davidbeechey, line 1: Move to `lib/sensors/src/optical_flow.rs` (you'll need to add it to `lib/sensors/src/lib.rs` too

optical_flow.rs
```diff
@@ -0,0 +1,127 @@
...
+
+#[derive(Serialize)]
+struct Payload {
+    x: i16,
+    y: i16,
+}
+
+
+// Make main function with flexible error handling
+pub fn main() -> Result<(), Box<dyn std::error::Error>> {
```
davidbeechey, line 92: This function won't live here, it'll be in one of the `board/{board_name}/tasks` folders as a task. Have a look at `boards/stm32l476rg/src/tasks/temperature.rs` as an example. (Just copy this code into a structure similar to that for now)

optical_flow.rs
```diff
@@ -0,0 +1,127 @@
...
+
+pub struct PMW3901<SPI> {
+    spi: SPI,
+
+}
+
+impl<SPI> PMW3901<SPI>
+where
+    SPI: HypedSpi,
+{
```
davidbeechey, line 24:
```suggestion
impl<SPI: HypedSpi> PMW3901<SPI>
```
You should be able to do something shorter like this instead

optical_flow.rs
```diff
@@ -0,0 +1,127 @@
...
+
+
+// implement get_motion function first
+// add readByteFromRegister (?) function to HypedSPI trait which takes register address and gives data
+
+const REG_MOTION_BURST: u8 = 0x16;
+const TIMEOUT: Duration = Duration::from_secs(5);
+
+
+pub struct PMW3901<SPI> {
```
davidbeechey, line 16:
```suggestion
pub struct PMW3901<SPI: HypedSpi> {
```
It's also standard to have single uppercase characters as type parameters, like `T`. But I don't mind `SPI` for now

optical_flow.rs
```diff
@@ -0,0 +1,127 @@
...
+    SPI: HypedSpi,
+{
+
+    
+    pub fn new(spi: SPI) -> Self {
+        PMW3901 { spi }
+    }
+
+    // **ATTEMPTED PYTHON IMPLEMENTATION OF get_motion**
+    pub fn get_motion(&mut self) -> Result<(i16, i16), &'static str>{
```
davidbeechey, line 32: You'll want to define an error enum type like `OpticalFlowError` instead of returning a string here. Have a look at `TemperatureError` as an example (you'll probably want to have an enum variant for `SpiError` for example

optical_flow.rs
davidbeechey: Probs better to call this `OpticalFlow` also

davidbeechey (changes requested):

davidbeechey:
Moved to #48.

PR #33, opened 2024-10-14
LOC - Implement Navigator

misha7b:
TODO:

- [ ] Implement Navigator.cpp from 2024 codebase in Rust.
- [ ] Think of a new name. "Localisator"??


lib/localisation/Cargo.toml
```diff
@@ -5,3 +5,4 @@ edition = "2021"
 
 [dependencies]
 nalgebra = "0.33.0"
```
davidbeechey, line 7:
```suggestion
nalgebra = { version = "0.33.0", default-features = false }
```
See: [https://nalgebra.org/docs/user_guide/wasm_and_embedded_targets/#compiling-without-standard-library](https://nalgebra.org/docs/user_guide/wasm_and_embedded_targets/#compiling-without-standard-library)

davidbeechey (changes requested):

PR #35, opened 2024-10-15
INFRA - Add STM32F767ZI board support

davidbeechey:
Don't really need this, but I had to create it to test something anyway so might ask well. Proves that it's super easy to support more than one board - didn't have to change any logic at all.


TomLonergan03 (approved):
assuming this works, lgtm

PR #36, opened 2024-10-16
LOC - Accelerometer Preprocessor

Sean-Leishman:
Migrates accelerometer preprocessor implementation from Hyped-2024


lib/core/src/types.rs
davidbeechey, line 1: If these types are only going to be used by the localisation code then they should go in `lib/localisation/src/types.rs`. (Core should be for things that other crates are likely to use as well)

davidbeechey: Actually, even just having them in the accelerometer preprocessing file is probably ok

lib/localisation/src/preprocessing/accelerometer.rs
```diff
@@ -1 +1,185 @@
-//test
+use heapless::Vec;
+use hyped_core::types::{AccelerometerData, RawAccelerometerData, SensorChecks};
+use hyped_core::types::{K_NUM_ACCELEROMETERS, K_NUM_ALLOWED_ACCELEROMETER_OUTLIERS, K_NUM_AXIS};
+
+pub struct AccelerometerPreprocessor {
+    num_reliable_accelerometers_: i32,
```
davidbeechey: I think the trailing underscore is a C++ thing so you can remove them

lib/core/src/types.rs
```diff
@@ -14,6 +16,28 @@ impl DigitalSignal {
...
+    x: i32,
+    y: i32,
+    z: i32,
+    timestamp: i32,
+    is_sensor_active: bool,
+}
+
+pub const K_NUM_ACCELEROMETERS: usize = 3;
+pub const K_NUM_AXIS: usize = 3;
+pub const K_NUM_ALLOWED_ACCELEROMETER_OUTLIERS: i32 = 2;
```
davidbeechey: Why a mix of `usize` and `i32`?

Sean-Leishman: Updated to use K_NUM_ALLOWED_ACCELEROMETER_OUTLIERS with `usize`. The mix is used so as to remain concise with initializing the required vectors with K_NUM_* fairly succiently

lib/core/src/types.rs
Sean-Leishman: Moved into a lib/localisation/src/types.rs just for now in case they are required for other files (such as filters)

lib/localisation/src/preprocessing/accelerometer.rs
```diff
@@ -1 +1,300 @@
...
+    lower_bound: f32,
+    upper_bound: f32,
+}
+
+impl Quartiles {
+    pub fn new(q1: f32, q2: f32, q3: f32, is_unreliable: bool) -> Self {
+        let mut bound_factor = 1.5;
+        if is_unreliable {
+            bound_factor = 1.2;
+        }
```
TomLonergan03:
```rust
let bound_factor = if is_unreliable {
    1.2
} else {
    1.5
};
```
is nicer imo

lib/localisation/src/preprocessing/accelerometer.rs
```diff
@@ -1 +1,300 @@
...
+    num_reliable_accelerometers: i32,
+    reliable_accelerometers: [bool; K_NUM_ACCELEROMETERS],
+    num_outliers_per_accelerometer: [i32; K_NUM_ACCELEROMETERS],
+}
+
+impl Default for AccelerometerPreprocessor {
+    fn default() -> Self {
+        Self::new()
+    }
+}
```
TomLonergan03: what is this for?

lib/localisation/src/preprocessing/accelerometer.rs
```diff
@@ -1 +1,300 @@
...
+    }
+
+    pub fn handle_outliers(
+        &mut self,
+        data: AccelerometerData<K_NUM_ACCELEROMETERS>,
+    ) -> Option<AccelerometerData<K_NUM_ACCELEROMETERS>> {
+        let quartiles = match self.calculate_quartiles(&data) {
+            Some(quartiles) => quartiles,
+            None => return None,
+        };
```
TomLonergan03:
```rust
let quartiles = self.calculate_quartiles(&data)?;
```
`?` operator will give back the wrapped value of a `Some(...)` or early return if `None`

lib/localisation/src/preprocessing/accelerometer.rs
```diff
@@ -1 +1,300 @@
...
+            const SIZE: usize = K_NUM_ACCELEROMETERS - 1;
+            let mut filtered_data: AccelerometerData<SIZE> =
+                AccelerometerData::from_slice(&[0.0; SIZE]).unwrap();
+            let mut filtered_data_idx = 0;
+            data.iter().enumerate().for_each(|(i, val)| {
+                if self.reliable_accelerometers[i] {
+                    filtered_data[filtered_data_idx] = *val;
+                    filtered_data_idx += 1;
+                }
+            });
```
TomLonergan03:
```rust
let filtered_data: AccelerometerData<SIZE> = data
    .iter()
    .enumerate()
    .filter(|(i, _)| self.reliable_accelerometers[*i])
    .map(|(_, &val)| val)
    .collect();
```
this might be nicer (if i understand correctly what you're doing) (and if i remember how rust works)

lib/localisation/src/preprocessing/accelerometer.rs
```diff
@@ -1 +1,300 @@
...
+            axis.iter().for_each(|val| {
+                magnitude += val * val;
+            });
+            accelerometer_data[i] = magnitude.sqrt();
+        });
+
+        let clean_accelerometer_data = match self.handle_outliers(accelerometer_data) {
+            Some(data) => data,
+            None => return None,
+        };
```
TomLonergan03:
```rust
let clean_accelerometer_data = self.handle_outliers(accelerometer_data)?;
```

TomLonergan03 (commented):
looks good, here are some suggestions
(i don't know much about what you're trying to do so they may be wrong)

lib/localisation/Cargo.toml
```diff
@@ -5,3 +5,6 @@ edition = "2021"
 
 [dependencies]
 nalgebra = "0.33.0"
```
davidbeechey:
```suggestion
nalgebra = { version = "0.33.0", default-features = false }
```
See: [https://nalgebra.org/docs/user_guide/wasm_and_embedded_targets/#compiling-without-standard-library](https://nalgebra.org/docs/user_guide/wasm_and_embedded_targets/#compiling-without-standard-library)

davidbeechey (changes requested):

lib/localisation/src/preprocessing/accelerometer.rs
Sean-Leishman: Left over from a prior impl

lib/core/src/types.rs
davidbeechey: Can all of the localisation stuff be removed from this file now then?

lib/localisation/src/preprocessing/accelerometer.rs
```diff
@@ -1 +1,283 @@
...
+                .enumerate()
+                .filter(|(i, _)| self.reliable_accelerometers[*i])
+                .map(|(_, val)| *val)
+                .collect();
+            quartiles = self.get_quartiles(&filtered_data);
+        } else {
+            return None;
+        }
+
+        Some(quartiles)
```
davidbeechey: Might be nicer here to return early with `Some(quartiles)` rather than have an uninitialised `quartiles` variable. Then you'd not need the last `else` and can just return `None`

davidbeechey (changes requested):
Looks good (I'm not familiar with the logic, but the code looks okay at least). 

Please add Rust doc comments (`///`) for `Quartiles` and `AccelerometerPreprocessor` explaining what they do in reasonable detail (at least a few lines), and comments on the individual functions too. Thanks!

lib/localisation/src/preprocessing/accelerometer.rs
```diff
@@ -1 +1,283 @@
...
+            RawAccelerometerData::from_slice(&[
+                Vec::from_slice(&[1.0, 2.0, 3.0]).unwrap(), // sqrt(14) = 3.74
+                Vec::from_slice(&[4.0, 5.0, 6.0]).unwrap(), // sqrt(77) = 8.77
+                Vec::from_slice(&[7.0, 8.0, 9.0]).unwrap(), // sqrt(194) = 13.93
+                Vec::from_slice(&[10.0, 11.0, 12.0]).unwrap(), // sqrt(365) = 19.1
+            ])
+            .unwrap();
+
+        let processed_data = preprocessor.process_data(raw_data);
+        assert!(processed_data.is_some());
```
TomLonergan03, line 213: It would be good to have a test that checks more than `process_data` just returning `Some(_)`

In this test `processed_data == Some(Vec::from_slice(&[3.7416575, 8.774964, 13.928389, 19.104973]).unwrap());`, if those are the right values then you should assert them, otherwise work out what they should be

lib/localisation/src/preprocessing/accelerometer.rs
```diff
@@ -1 +1,283 @@
...
+            AccelerometerData::from_slice(&[0.0; K_NUM_ACCELEROMETERS]).unwrap();
+        let mut magnitude: f32 = 0.0;
+
+        data.iter().enumerate().for_each(|(i, axis)| {
+            magnitude = 0.0;
+            axis.iter().for_each(|val| {
+                magnitude += val * val;
+            });
+            accelerometer_data[i] = magnitude.sqrt();
+        });
```
TomLonergan03:
```suggestion        
        let accelerometer_data: AccelerometerData<K_NUM_ACCELEROMETERS> = data
            .iter()
            .map(|axis| axis.iter().fold(0.0, |acc, val| acc + val * val).sqrt())
            .collect();
```
I think this is a nicer way to do this, avoids unnecessary `mut`s
Maybe i'm just too functional pilled tho

TomLonergan03 (approved):
looks good, made a couple suggestions to make things a little more concise or robust

misha7b (approved):
Looks good!

davidbeechey (approved):

PR #38, opened 2024-10-18
SNS - Create `HypedSpi` trait

davidbeechey:
Creates the `HypedSpi` trait for abstracting SPI. Will enable a `MockSpi` struct and for SPI sensors to be implemented without having to wait for the STM32 implementation of SPI to be implemented first.

**To do:**
- [ ] Is the way that word size is implemented okay?

The alternative to this if word size is a problem is to just use the Embassy implementation of SPI to implement sensors that are board-specific. E.g. create optical flow in `boards/stm32h7.../sensors`. Would mean that if we want to support this sensor on another board we would have to duplicate the entire sensor implementation (and maybe change it), rather than just implementing `HypedSpi` for it. Thoughts?


lib/io/src/spi.rs
```diff
@@ -0,0 +1,22 @@
+/// SPI errors that can occur
+/// From: https://docs.embassy.dev/embassy-stm32/git/stm32f103c8/spi/enum.Error.html
+pub enum SpiError {
```
TomLonergan03, line 3: is there a reason to redefine this instead of doing `pub use embassy_stm32::spi::Error as SpiError;` (i think that's the right way to reexport the type)

lib/io/src/spi.rs
```diff
@@ -0,0 +1,22 @@
...
+/// A word is either u8 or u16
+pub enum WordSize {
+    U8(u8),
+    U16(u16),
+}
+
+/// SPI trait used to abstract SPI and allow for mocking
+pub trait HypedSpi {
+    /// Read data into `words` from the SPI sensor
+    fn read(&mut self, words: &mut [WordSize]) -> Result<(), SpiError>;
```
TomLonergan03, line 19: is the out-param the best way to return the read words?

lib/io/src/spi.rs
davidbeechey: Part of the bind we're having with abstractions is that we can't use `embassy_stm32` because we have to enable exactly one board feature for it which makes the crate board-specific (so we wouldn't be able to reuse it across different boards)

lib/io/src/spi.rs
davidbeechey: Honestly not sure, idk if we can just return a slice because we'd need to know its size in advance

licornes-fluos (approved):

TomLonergan03 (approved):

davidbeechey:
Might have to modify the `WordSize` later on, or just replace it with a generic type `T`, but it's good enough to start with.

PR #39, opened 2024-10-22
SNS - Create `MockI2c` (and temperature sensor tests)

davidbeechey:
Creates `MockI2c` for use in testing. Temperature sensor (STTS22H) basic tests also created to try it out (closes HYPE-53).

This changes the way `HypedI2c` is passed into `Temperature::new()`, it is now a mutable reference to the `HypedI2c` instance.


lib/io/src/i2c.rs
```diff
@@ -22,3 +21,40 @@ pub trait HypedI2c {
...
+
+    impl crate::i2c::HypedI2c for MockI2c {
+        /// Reads a byte by looking up the device address and register address in the map
+        fn read_byte(&mut self, device_address: u8, register_address: u8) -> Option<u8> {
+            self.values
+                .get(&(device_address, register_address))
+                .copied()
+        }
+
+        /// Always succeeds, but does nothing
```
TomLonergan03: something you could do here if you care is store the writes that occur and have a getter, then you could test that the correct writes occur

lib/io/src/i2c.rs
```diff
@@ -22,3 +21,40 @@ pub trait HypedI2c {
         data: u8,
     ) -> Result<(), I2cError>;
 }
+
+pub mod mock_i2c {
+    use heapless::FnvIndexMap;
+
+    /// A fixed-size map of I2C values, indexed by device address and register address
+    type I2cValues = FnvIndexMap<(u8, u8), u8, 16>;
```
TomLonergan03: can the return type be an `Option<u8>`? then you can test failure cases as well

lib/io/src/i2c.rs
TomLonergan03: and/or have a list of write results, so you can check failure cases again

TomLonergan03 (approved):
looks good, a couple suggestions for how you could make the mock more coveringer

H-Allen (approved):

PR #40, opened 2024-10-23
SNS - Pneumatics Control

davidbeechey:
Describes the valid states of the solenoids in the pneumatics system.

- Document describing this: https://docs.google.com/document/d/1qPipkJgf-3Wg8KbpnK8xv9aAjqgkh0oMshStEEHdcVQ/edit

**To do:**

Haven't decided if this should just describe the current state or if it should act as the central controller for this, which would be more complex...


lib/control/src/pneumatics.rs
```diff
@@ -0,0 +1,193 @@
...
+
+pub struct PneumaticsOutput {
+    pub yellow: DigitalSignal,
+    pub pink: DigitalSignal,
+    pub orange: DigitalSignal,
+    pub green: DigitalSignal,
+    pub blue_1: DigitalSignal,
+    pub blue_2: DigitalSignal,
+}
+
```
davidbeechey: This can all be deleted now I think

lib/control/src/pneumatics.rs
```diff
@@ -0,0 +1,193 @@
...
+struct Pneumatics<P: GpioOutputPin> {
+    brakes: BrakeState,
+    lateral_suspension: LateralSuspensionState,
+    brake_pin: P,
+    lateral_suspension_pin: P
+}
+
+impl<P: GpioOutputPin> Pneumatics<P> {
+    fn new(mut brake_pin: P, mut lateral_suspension_pin: P) -> Self {
+        brake_pin.set_high();
```
davidbeechey: The brakes are actually electrically failsafe, so they are engaged when not powered. (So set the pin to high to disengage and low to engage.) Would be good to document this behaviour in a Rust doc comment for `Pneumatics`

lib/control/src/pneumatics.rs
```diff
@@ -0,0 +1,193 @@
...
+
+    fn deploy_lateral_suspension(&mut self) {
+        self.lateral_suspension = LateralSuspensionState::Deployed;
+        self.lateral_suspension_pin.set_high();
+    }
+
+    fn retract_lateral_suspension(&mut self) {
+        self.lateral_suspension = LateralSuspensionState::Retracted;
+        self.lateral_suspension_pin.set_low();
+    }
```
davidbeechey: It'll probably be useful to have getters for the current brakes status and current lateral suspension status too

boards/stm32l476rg/src/tasks/pneumatics.rs
```diff
@@ -0,0 +1,11 @@
+use embassy_time::{Duration, Timer};
+
+#[embassy_executor::task]
+pub async fn control_pneumatics() -> ! {
+    let p = embassy_stm32::init(Default::default());
+    let mut pneumatics_control = PneumaticsControl::new();
```
davidbeechey: Just need to complete this by creating two GPIO inputs and trying to create the pneumatics control struct

lib/control/src/pneumatics.rs
```diff
@@ -0,0 +1,68 @@
...
+    Deployed,
+    Retracted,
+}
+
+struct Pneumatics<P: GpioOutputPin> {
+    brakes: BrakeState,
+    lateral_suspension: LateralSuspensionState,
+    brake_pin: P,
+    lateral_suspension_pin: P
+}
```
davidbeechey: I can't see this being a issue we encounter, but at the moment `brake_pin` and `lateral_suspension_pin` both have to be the same type `P`. Might be better to have two type parameters, one for each pin. @TomLonergan03 do you agree?

TomLonergan03: i wouldn't bother, but it makes not much difference. if i understand correctly that only matters if you end up with the pins each running on different gpio pin structs, which i doubt will happen. maybe in testing you'd want to connect one to a fake pin and one to a real one?

jpfbastos (approved):

boards/stm32l476rg/src/tasks/pneumatics.rs
```diff
@@ -0,0 +1,12 @@
+use embassy_time::{Duration, Timer};
+
+#[embassy_executor::task]
+pub async fn control_pneumatics<P: GpioOutputPin>(brake_gpio: P, lateral_suspension_gpio: P) -> ! {
+    let p = embassy_stm32::init(Default::default());
+    let pneumatics = Pneumatics::new(brake_gpio, lateral_suspension_gpio);
+    let mut pneumatics_control = PneumaticsControl::new(pneumatics);
```
TomLonergan03: what is `PneumaticsControl`?

TomLonergan03: yeah this whole file isn't included in `tasks.rs` so it doesn't get built. i'm pretty sure there isn't a `PneumaticsControl` struct, and @davidbeechey is there a warning that can fail the build action for an unincluded file? the language server seems to know at least

davidbeechey: doesn't look like there's any standard tools for checking for unincluded files

TomLonergan03 (approved):

H-Allen (approved):

PR #14, opened 2024-09-13
SNS - I2C Mux (HYPE-13)

davidbeechey:
Translates the I2C mux code from last year's repository. This might not be the best interface for the mux, so it might change later once we have a better idea about how it will be used.

~(This PR cherry-picks the I2C bits from PR #12)~

PR #41, opened 2024-10-29
SNS - ADC Mux

Vlad118:


davidbeechey (changes requested):
Looks good, would be good to have a function for just reading a single channel of the mux. To do so it'd be best to have the channel addresses as an enum, so that'll require modification to the `read_data` function too.

**Edit:** actually, it would be really good if we had something like `AdcMuxChannel` which implemented the `HypedAdc` trait from PR #41. That would mean we don't need to care whether an ADC sensor is plugged into the ADC on the ST board or into the mux, the sensor can just use anything that implements `HypedAdc`. It would need to take in a shared I2C instance (using `Mutex` like in the I2C mux PR #14), the mux address and the channel address.

lib/io/src/adc_mux.rs
```diff
@@ -0,0 +1,56 @@
...
+    AdcChannel26 = 0x26, 
+    AdcChannel27 = 0x27,
+}
+
+/// Possible errors that can occur when interacting with the ADC Mux
+#[derive(Debug)]
+pub enum MuxError {
+    I2cError(I2cError),
+    TooManyFaultyChannels,
+    TooManyChannels,
```
davidbeechey: I don't think we need these anymore

lib/io/src/adc_mux.rs
```diff
@@ -0,0 +1,56 @@
...
+const ADC_MUX_ADDRESS: u8 = 0x1D;
+pub enum AdcChannelAddress {
+    AdcChannel20 = 0x20, 
+    AdcChannel21 = 0x21, 
+    AdcChannel22 = 0x22, 
+    AdcChannel23 = 0x23, 
+    AdcChannel24 = 0x24, 
+    AdcChannel25 = 0x25, 
+    AdcChannel26 = 0x26, 
+    AdcChannel27 = 0x27,
```
davidbeechey: wondering maybe these should just be called `AdcChannelN` instead, because I don't think we're actually gonna care about the address

boards/stm32l476rg/src/io.rs
```diff
@@ -1,3 +1,5 @@
 pub mod adc;
 pub mod gpio;
 pub mod i2c;
+pub mod adc;
```
davidbeechey:
```suggestion
```

davidbeechey (changes requested):
Few changes needed + update the comments accordingly

davidbeechey (approved):

TomLonergan03 (approved):
you might need to make some changes in this depending what happens with #49

H-Allen (approved):

PR #43, opened 2024-10-31
LOC - Implement Kalman Filter (HYPE-46)

misha7b:
Implementing the Kalman Filter from the 2024 codebase.
- Changed the equations to a more common form
- Added a test


lib/localisation/src/filtering/kalman_filter.rs
```diff
@@ -0,0 +1,144 @@
...
+    control_matrix: DMatrix<f64>,
+
+    // Observation matrix
+    observation_matrix: DMatrix<f64>,
+
+    // Process noise covariance
+    process_noise: DMatrix<f64>,
+
+    // Measurement noise covariance
+    measurement_noise: DMatrix<f64>,
```
davidbeechey: Use `///` here for Rust doc to work, plus might look better without the spaces?

lib/localisation/src/filtering/kalman_filter.rs
```diff
@@ -0,0 +1,144 @@
+use nalgebra::{DMatrix, DVector};
+
+pub struct KalmanFilter {
```
davidbeechey, line 10: Would be good to write a Rust doc comment briefly explaining what the Kalman Filter is doing and why we use it

lib/localisation/src/filtering/kalman_filter.rs
```diff
@@ -0,0 +1,144 @@
...
+            covariance: initial_covariance,
+            transition_matrix,
+            observation_matrix,
+            control_matrix,
+            process_noise,
+            measurement_noise,
+        }
+    }
+
+    // Predict step
```
davidbeechey:
```suggestion
    /// Predict step
```
Might be good also to quickly explain what it's doing to predict (what is the prediction based on?)

lib/localisation/src/filtering/kalman_filter.rs
```diff
@@ -0,0 +1,144 @@
...
+        // x_k = F * x_k-1 + B * u_k
+        self.state = &self.transition_matrix * &self.state + &self.control_matrix * control_input;
+
+        // P_k = F * P_k-1 * F^T + Q
+        self.covariance =
+            &self.transition_matrix * &self.covariance * self.transition_matrix.transpose()
+                + &self.process_noise;
+    }
+
+    // Update step
```
davidbeechey: same here

davidbeechey (changes requested):
Looks good. Slight problem around `no_std`, `nalgebra` links to `libstd` by default (which defeats the point of being `no_std`). Pretty sure this is a problem, but should be other ways to do this. (See [https://nalgebra.org/docs/user_guide/wasm_and_embedded_targets/#for-embedded-development](https://nalgebra.org/docs/user_guide/wasm_and_embedded_targets/#for-embedded-development).) We can have a look next week

lib/localisation/src/filtering/kalman_filter.rs
```diff
@@ -0,0 +1,144 @@
...
+        let innovation_covariance =
+            &self.observation_matrix * &self.covariance * self.observation_matrix.transpose()
+                + &self.measurement_noise;
+
+        // K = P_k * H^T * S^-1
+        let kalman_gain = &self.covariance
+            * self.observation_matrix.transpose()
+            * innovation_covariance
+                .try_inverse()
+                .expect("Failed to invert innovation covariance matrix");
```
TomLonergan03: do you know that the inverse will always succeed here?

lib/localisation/src/filtering/kalman_filter.rs
```diff
@@ -0,0 +1,164 @@
+use nalgebra::{Matrix2, Vector1, Vector2};
```
davidbeechey: Sorry gonna be really picky and ask for a space between the import and the start of the comment

davidbeechey (approved):

lib/localisation/src/filtering/kalman_filter.rs
```diff
@@ -0,0 +1,164 @@
...
+        // Calculate the inverse of the innovation covariance
+
+        let a = innovation_covariance[(0, 0)];
+        let b = innovation_covariance[(0, 1)];
+        let c = innovation_covariance[(1, 0)];
+        let d = innovation_covariance[(1, 1)];
+
+        let determinant = a * d - b * c;
+
+        let innovation_covariance_inv = Matrix2::new(d, -b, -c, a) / determinant;
```
TomLonergan03: is the determinant definitely never going to be 0? if it is 0 this panics, if you can guarantee it isn't then there isn't a need to manually do the inverse instead of using the lib function + `unwrap()` (unless there is and i'm unaware)

TomLonergan03 (changes requested):

lib/localisation/src/filtering/kalman_filter.rs
TomLonergan03: add a comment that explains why the determinant will never be 0 and thus the inverse cannot fail

TomLonergan03 (approved):

PR #47, opened 2024-11-11
LOC - Implement Localizer

misha7b:
Transfer navigator.cpp from the 2024 codebase.

- [x] Think of new name.  "Localisator"?


kshxtij:
Localizer?

lib/localisation/src/control/localizer.rs
```diff
@@ -0,0 +1,186 @@
...
+    types::{RawAccelerometerData, K_NUM_ACCELEROMETERS, K_NUM_AXIS},
+};
+
+use heapless::Vec;
+
+use libm::pow;
+use nalgebra::{Matrix2, Vector1, Vector2};
+
+const DELTA_T: f64 = 0.01;
+const STRIPE_WIDTH: f64 = 1.0;
```
davidbeechey: would be good to add a comment with the units for these

lib/localisation/src/control/localizer.rs
```diff
@@ -0,0 +1,186 @@
...
+
+        let keyence_status = self
+            .keyence_checker
+            .check_keyence_agrees(keyence_data.clone());
+
+        if keyence_status == SensorChecks::Unacceptable {
+            //TODOLater: Change state
+            return;
+        } else {
+            //TODOLater: Check unit of keyence data
```
davidbeechey: what unit is this referring to?

lib/localisation/src/control/localizer.rs
```diff
@@ -0,0 +1,186 @@
...
+    ) {
+        let processed_optical_data = process_optical_data(optical_data.clone());
+        self.optical_val = processed_optical_data as f64;
+
+        let keyence_status = self
+            .keyence_checker
+            .check_keyence_agrees(keyence_data.clone());
+
+        if keyence_status == SensorChecks::Unacceptable {
+            //TODOLater: Change state
```
davidbeechey: rather than handling the logic for changing state in here, the whole `preprocessor` function could return `Result<(), PreprocessorError>` or similar in this case so that the caller can deal with it instead

lib/localisation/src/control/localizer.rs
```diff
@@ -0,0 +1,186 @@
...
+        } else {
+            //TODOLater: Check unit of keyence data
+            self.keyence_val = keyence_data[0] as f64;
+        }
+
+        let mut accelerometer_preprocessor = AccelerometerPreprocessor::new();
+        let processed_accelerometer_data =
+            accelerometer_preprocessor.process_data(accelerometer_data);
+        if processed_accelerometer_data.is_none() {
+            // TODOLater: Change state
```
davidbeechey: likewise here

lib/localisation/src/control/localizer.rs
```diff
@@ -0,0 +1,186 @@
...
+        &mut self,
+        optical_data: Vec<f64, 2>,
+        keyence_data: Vec<u32, 2>,
+        accelerometer_data: RawAccelerometerData<K_NUM_ACCELEROMETERS, K_NUM_AXIS>,
+    ) {
+        self.preprocessor(
+            optical_data.clone(),
+            keyence_data.clone(),
+            accelerometer_data.clone(),
+        );
```
davidbeechey: if you change the return type then `iteration` will also return a `Result` type, so just pass the error up

lib/localisation/src/control/localizer.rs
```diff
@@ -0,0 +1,186 @@
...
+        //TODOLater: Check unit of keyence data
+        let measurement = Vector2::new(self.keyence_val * STRIPE_WIDTH, self.optical_val);
+
+        self.kalman_filter.update(&measurement);
+
+        let state = self.kalman_filter.get_state();
+
+        self.displacement = state[0];
+        self.velocity = state[1];
+        self.acceleration = (self.velocity - self.previous_velocity) / DELTA_T; //TODOLater: is this accurate enough?
```
davidbeechey: How are you going to determine whether this is accurate enough? Is that something that needs physical testing?

lib/localisation/src/control/localizer.rs
```diff
@@ -0,0 +1,186 @@
...
+
+        if keyence_status == SensorChecks::Unacceptable {
+            //TODOLater: Change state
+            return;
+        } else {
+            //TODOLater: Check unit of keyence data
+            self.keyence_val = keyence_data[0] as f64;
+        }
+
+        let mut accelerometer_preprocessor = AccelerometerPreprocessor::new();
```
davidbeechey: Might be better to just have one instance of `AccelerometerPreprocessor` like with Keyence so it's not being constantly created

lib/localisation/src/control/localizer.rs
```diff
@@ -0,0 +1,186 @@
...
+        let initial_state = Vector2::new(0.0, 0.0);
+        let initial_covariance = Matrix2::new(1.0, 0.0, 0.0, 1.0);
+        let transition_matrix = Matrix2::new(1.0, DELTA_T, 0.0, 1.0);
+        let control_matrix = Vector2::new(0.5 * DELTA_T * DELTA_T, DELTA_T);
+        let observation_matrix = Matrix2::new(1.0, 0.0, 0.0, DELTA_T);
+
+        // Assuming frequency of 6400hz for IMU at 120 mu g / sqrt(Hz)
+        // standard deviation = 120 * sqrt(6400) = 9600 mu g = 0.0096 g
+        //                    = 0.0096 * 9.81 = 0.094176 m/s^2
+        // variance =  0.094176^2 = 0.0089 m/s^2
```
davidbeechey, line 48: really good comments here

lib/localisation/src/control/localizer.rs
```diff
@@ -0,0 +1,186 @@
...
+    }
+}
+
+impl Default for Localizer {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl Localizer {
```
davidbeechey, line 102: Some Rust doc `///` comments for `Localizer` and the `preprocessor` and `iteration` functions would be great

lib/localisation/src/preprocessing/optical.rs
```diff
@@ -2,20 +2,15 @@ use heapless::Vec;
 use libm::sqrtf;
 
 /// Processes the raw optical data to get the magnitude and added to the optical data for each sensor
-pub fn process_data(raw_optical_data: Vec<Vec<f64, 2>, 2>) -> Vec<f32, 2> {
-    let mut optical_data: Vec<f32, 2> = Vec::from_slice(&[0.0, 0.0]).unwrap();
+pub fn process_optical_data(raw_optical_data: Vec<f64, 2>) -> f32 {
+    let mut magnitude: f32 = 0.0;
```
davidbeechey: Is it possible to keep this generic to any number of optical flow sensors? Could take the number as a constant `NUM_OPTICAL_FLOW_SENSORS` like with accelerometers, and just have a `Vec` of length 1 if there's only one?

lib/localisation/src/control/localizer.rs
```diff
@@ -0,0 +1,186 @@
...
+
+    #[test]
+    fn test_localizer_with_zeros() {
+        let mut localizer = Localizer::default();
+
+        let optical_data: Vec<f64, 2> = Vec::from_slice(&[0.0, 0.0]).unwrap();
+
+        let raw_keyence_data: Vec<u32, 2> = Vec::from_slice(&[0, 0]).unwrap();
+
+        let raw_accelerometer_data: RawAccelerometerData<K_NUM_ACCELEROMETERS, K_NUM_AXIS> =
```
davidbeechey: Didn't spot this earlier, but we don't need `K_` prefix for `K_NUM_ACCELEROMETERS` or `K_NUM_AXIS`

davidbeechey (changes requested):
Looking good, just a few things to change and comments!

lib/localisation/src/control/localizer.rs
misha7b: If it is in metres

lib/localisation/src/control/localizer.rs
misha7b: Forgot to change it back, this should have just been the accelerometer sensor value

lib/localisation/src/control/localizer.rs
davidbeechey: Are we able to make this configurable? Ideally we'd have a constant somewhere for `KEYENCE_STRIPE_DISTANCE` so that we can change whether the posts are 1m, 1.5m, etc. apart on the fly

TomLonergan03 (approved):

davidbeechey (approved):

PR #49, opened 2024-11-16
SNS - Shared IO

davidbeechey:
Wraps all IO that could be shared (currently only I2C I think?) in `Mutex<RefCell<T>>` so that the same `HypedI2c` instance can be used for multiple sensors.


lib/sensors/src/temperature.rs
```diff
@@ -32,20 +40,25 @@ impl<'a, T: HypedI2c> Temperature<'a, T> {
...
+            Some(byte) => byte,
+            None => {
+                return None;
+            }
+        };
+
+        let temperature_low_byte = match self.i2c.lock(|i2c: &RefCell<T>| {
+            i2c.borrow_mut()
+                .read_byte(self.device_address, STTS22H_DATA_TEMP_L)
+        }) {
```
TomLonergan03: is it possible to move the `Mutex<Refcell<...>>` into the `i2c` impl so this is done there instead of needing done on every call in every sensor? maybe wrapping the `i2c` field of `Stm32f767ziI2c` in the `Mutex<RefCell<...>>` or something similar?

TomLonergan03 (approved):

lib/sensors/src/temperature.rs
davidbeechey: Good shout, I've changed this. Still a bit clunky because we still have to pass in a mutable reference to the sensors to allow us to read `MockI2c` values afterwards

boards/stm32l476rg/src/io/i2c.rs
```diff
@@ -45,7 +53,7 @@ impl<'d> HypedI2c for Stm32l476rgI2c<'d> {
 
 impl<'d> Stm32l476rgI2c<'d> {
     /// Create a new instance of our I2C implementation for the STM32L476RG
-    pub fn new(i2c: I2c<'d, Blocking>) -> Self {
+    pub fn new(i2c: Mutex<CriticalSectionRawMutex, RefCell<I2c<'d, Blocking>>>) -> Self {
```
TomLonergan03: can you also just take `I2c<'d, Blocking>` here, and mutex refcell it here?
Then the whole sharing system is completely transparent to the interface user

but i can see that enforcing this in the interface makes it clearer to future i2c implementers that this kind of thing needs to happen

TomLonergan03: actually this isn't a trait method so i'd do the mutexing inside the constructor if that's possible

davidbeechey: I don't think so because it'd then own that I2C?

TomLonergan03: yeah, and pass around the `hyped_i2c` instead? or is that not how it works currently

davidbeechey: but then `hyped_i2c` would have to be wrapped in the mutex if it's being passed into multiple things?

jpfbastos (approved):

H-Allen (approved):

boards/stm32l476rg/src/io/i2c.rs
TomLonergan03: do you have a cloneable `Arc` equivalent for embassy? otherwise yeah you're right

PR #50, opened 2024-11-16
MLC - CanOpen

jpfbastos:


lib/motors/src/can_open_message.rs
```diff
@@ -0,0 +1,95 @@
...
+
+    pub const ENTER_OPERATIONAL_STATE: CanOpenMessage = CanOpenMessage {
+        id: 0x000,
+        command: 0x01,
+        index: 0x0000,
+        sub_index: 0x00,
+        data: 0x00000000,
+    };
+
+    pub const SET_FREQUENCY: CanOpenMessage = CanOpenMessage {
```
davidbeechey, line 59: possibly worth a comment explaining that `data` will be overwritten somewhere else

lib/motors/src/can_processor.rs
davidbeechey, line 1: should this be called `can_open_processor`?

davidbeechey (changes requested):
Looking good! Have made a few comments, but apart from that some good Rust doc comments would be great.

I'm tempted to say we should split this into CanOpen and motor controller stuff - let me know what you think :)

jpfbastos:
> Looking good! Have made a few comments, but apart from that some good Rust doc comments would be great.
> 
> I'm tempted to say we should split this into CanOpen and motor controller stuff - let me know what you think :)

yeah fair enough - i kinda wung it, and then forgot to a) add RustDoc comments and b) split the PRs

davidbeechey (approved):
Looks good!

lib/motors/src/can_open_processor.rs
```diff
@@ -0,0 +1,105 @@
...
+
+/// Convert a HypedCanFrame to a CanOpenMessage
+impl From<HypedCanFrame> for CanOpenMessage {
+    fn from(frame: HypedCanFrame) -> Self {
+        CanOpenMessage {
+            id: frame.can_id,
+            command: frame.data[0],
+            index: u16::from(frame.data[1]) | (u16::from(frame.data[2]) << 8),
+            sub_index: frame.data[3],
+            data: u32::from(frame.data[4])
```
TomLonergan03:
```rs
u32::from_le_bytes([frame.data[4], frame.data[5], frame.data[6], frame.data[7]])
```

(you can also do something like
```rs
let message_data = msg.data.to_le_bytes();
data.split_at_mut(4).1.copy_from_slice(&message_data);
```
for the other way round, but that's kinda ugly imo)

lib/motors/src/can_open_processor.rs
```diff
@@ -0,0 +1,105 @@
+use crate::can_open_message::{config_messages, messages, CanOpenMessage};
+use hyped_can::{CanError, HypedCan, HypedCanFrame};
+
+/// All types of messages that can be sent to the motor controller
+pub enum MessagesEnum {
```
TomLonergan03: just call it `Messages`

TomLonergan03 (changes requested):

lib/motors/src/can_open_processor.rs
TomLonergan03, line 1: another thing i just thought of, write a test that asserts `From<CanOpenMessage>` and `From<CanFrame>` are actually inverses of each other

lib/motors/src/can_open_processor.rs
jpfbastos: I changed for the first one, but I think I might keep what I have for the other way round, bc I think it's more readable (thanks for letting me know tho)

TomLonergan03 (approved):

PR #59, opened 2024-12-16
SNS - Sensor Bounds Checking

davidbeechey:
Adds a new `SensorValueRange` enum for wrapping sensor values so that we are forced to handle values that are in warning and/or critical range. At some point we will trigger a state transition when we receive a critical sensor reading, and probably have a higher log level message for warning values (and report this to telemetry).

I've also added a quick `build` script which builds the root workspace and all board crates too.

(Yes, I am procrastinating...)

**To Do:**
- [ ] ~Read sensor bounds from config file so that they can also be read by Telemetry~ Might be a lot of work to do on this PR, so since the logic for creating the bounds function is determined by the instantiating code we can deal with this separately on another PR


lib/sensors/src/lib.rs
```diff
@@ -2,3 +2,13 @@
 
 pub mod keyence;
 pub mod temperature;
+
+#[derive(PartialEq, Debug)]
```
TomLonergan03, line 7: maybe `#[must_use]`?

build
TomLonergan03, line 1: you could do this as a makefile instead, but may not be worth it
i just like makefiles lol

lib/sensors/src/temperature.rs
```diff
@@ -10,27 +12,37 @@ use hyped_i2c::{HypedI2c, I2cError};
...
     /// Create a new instance of the temperature sensor and attempt to configure it
+    ///
+    /// # Arguments
+    /// * `i2c` - The I2C peripheral to communicate with the temperature sensor
+    /// * `device_address` - The I2C address of the temperature sensor
+    /// * `calculate_bounds` - An optional function to override the default bounds calculation
     pub fn new(
         i2c: &'a mut T,
         device_address: TemperatureAddresses,
+        calculate_bounds: Option<fn(f32) -> SensorValueRange<f32>>,
```
TomLonergan03: i kinda don't like optional args, when would you want to use it?
my guess would be rename this to `new_with_bounds` and you can then have a `new` function that wraps that and passes in the default bounds

TomLonergan03 (approved):
hope ur revision is going well!

lib/sensors/src/temperature.rs
TomLonergan03: reason is, constructors that end up being called with `new(a, b, Some(5), None, None, Some(c))` are ugly and hard to read. one pretty solution for big things like that is builders

lib/sensors/src/lib.rs
davidbeechey: hadn't heard of this before - nice! will add

lib/sensors/src/temperature.rs
davidbeechey: yup good point, I like that better

davidbeechey: the rationale for supporting different bounds is for sensors like temp where the bounds might be different depending on where the sensor is on the pod

lib/sensors/src/lib.rs
TomLonergan03: i believe it's used for `Result`s and a few other things in the stdlib

lib/sensors/src/temperature.rs
```diff
@@ -24,13 +27,34 @@ impl<'a, T: HypedI2c> Temperature<'a, T> {
...
                 i2c,
                 device_address,
+                calculate_bounds: default_calculate_bounds,
+            }),
+            Err(e) => Err(TemperatureError::I2cError(e)),
+        }
+    }
+
+    /// Create a new instance of the temperature sensor with the specified bounds and attempt to configure it
+    pub fn new_with_bounds(
```
TomLonergan03, line 28: what i meant for this is you have `new_with_bounds`, which takes a required bounds function:
```rs
    /// Create a new instance of the temperature sensor with the specified bounds and attempt to configure it
    pub fn new_with_bounds(
        i2c: &'a mut T,
        device_address: TemperatureAddresses,
        calculate_bounds: fn(f32) -> SensorValueRange<f32>,
    ) -> Result<Self, TemperatureError> {
        // Set up the temperature sensor by sending the configuration settings to the STTS22H_CTRL register
        let device_address = device_address as u8;
        match i2c.write_byte_to_register(device_address, STTS22H_CTRL, STTS22H_CONFIG_SETTINGS) {
            Ok(_) => Ok(Self {
                i2c,
                device_address,
                calculate_bounds,
            }),
            Err(e) => Err(TemperatureError::I2cError(e)),
        }
    }
```
and then `new` just needs to be:
```rs
    /// Create a new instance of the temperature sensor and attempt to configure it
    pub fn new(
        i2c: &'a mut T,
        device_address: TemperatureAddresses,
    ) -> Result<Self, TemperatureError> {
        Self::new_with_bounds(i2c, device_address, default_calculate_bounds)
    }
```
which means no code duplication and no optional args anywhere

davidbeechey: ah crap, optional still being there was just silly copy+paste mistake lmao

build
platinumxy, line 1: Would it not be better to use a python script, as this will work on Linux and MacOS however not windows. we could do something like 
```python
import os
import subprocess

def build(path: str) -> None:
    print(f"Building {path}")
    subprocess.run(["cargo", "build"], cwd=path, check=True)

def main() -> None:
    # Build the main workspace
    print("Building main workspace")
    build(".")

    boards_dir = "boards"
    for dir_name in os.listdir(boards_dir):
        dir_path = os.path.join(boards_dir, dir_name)
        if os.path.isdir(dir_path):
            build(dir_path)

if __name__ == "__main__":
    main()
```

platinumxy (approved):

davidbeechey:
**To Do:**
- Read sensor bounds from config file so that they can also be read by Telemetry

H-Allen (approved):

PR #60, opened 2025-01-10
INFRA - Use `Watch` primitive for publishing sensor values

davidbeechey:
Use the `Watch` primitive from `embassy_sync` to hold the (single) latest value from a sensor, which can be updated from an individual task and awaited/polled for change from the spawning task. Allows us to properly use the sensor values in the main logic for that board rather than just printing their value from their tasks.

Also updates `read_temp` and `read_keyence` to take in an `i2c_bus` and `gpio_pin` respectively rather than them initialising the Embassy peripherals themselves (preventing them being used together). Might add an example later for reading temperature and keyence in one task.

**Notes:**
- Branch based on #14 for some changes to shared IO types.
- Bumps the version of `embassy_sync` to the latest from GitHub to get access to `embassy_sync::watch`.


boards/stm32l476rg/src/bin/temp_test.rs
TomLonergan03, line 1: `temperature_test.rs` cos "temp" makes me think "temporary"

boards/stm32l476rg/src/tasks/read_temp.rs
TomLonergan03, line 1: same on naming here

lib/io/hyped_i2c/hyped_i2c_derive/src/lib.rs
```diff
@@ -56,11 +56,29 @@ fn impl_hyped_i2c(ast: &syn::DeriveInput) -> TokenStream {
                     }),
                 }
             }
+
+            /// Write a byte to a device
+            fn write_byte(&mut self, device_address: u8, data: u8) -> Result<(), I2cError> {
+                let result = self.i2c.lock(|i2c| {
+                    i2c.borrow_mut().blocking_write(device_address, [data].as_ref())
```
TomLonergan03: this is a totally random aside, but why do you use blocking io instead of async?

TomLonergan03 (approved):
lgtm

lib/io/hyped_i2c/hyped_i2c_derive/src/lib.rs
davidbeechey: This is a good question, was wondering this myself last week. Will have a look at it during the hackathon maybe to see what advantages/disadvantages there are (I think we should probably be using async everywhere and looking at how we create executers etc. down the line)

davidbeechey: I've created an issue on Linear to have a look at this later: [HYPE-89/look-at-use-of-blocking-vs-async-io](https://linear.app/hyp-ed/issue/HYPE-89/look-at-use-of-blocking-vs-async-io).

Aux1r (approved):
Looks good to me, good commenting 👍

PR #63, opened 2025-01-23
MLC - CAN IO implementation

jpfbastos:


lib/io/hyped_can/hyped_can_derive/src/lib.rs
```diff
@@ -0,0 +1,67 @@
...
+                    Ok(_) => Some(read[0]),
+                    Err(_) => None,
+                }
+            }
+
+            /// Write a CAN frame to the CAN bus
+            fn write_frame(&mut self, message: &CanFrame) -> Result<(), CanError> {
+
+                let result = self.can.lock(|can| {
+                    can.borrow_mut().write()
```
davidbeechey: should be passing in the frame to write here (should probably rename `message` to `frame` too)

lib/io/hyped_can/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+    can_id: u32,   // 32 bit CAN_ID + EFF/RTR/ERR flags
+    data: [u8; 8], // data that is sent over CAN, split into bytes
+}
+
+/// CAN trait used to abstract the CAN operations
+pub trait HypedCan {
+    async fn read_frame(&mut self) -> Option<u8>;
+    fn try_read_frame(&mut self) -> Option<u8>;
+    async fn write_frame(&mut self, message: &CanFrame) -> Result<(), CanError>;
+    fn try_write_frame(&mut self, message: &CanFrame) -> Result<(), CanError>;
```
davidbeechey: Brief Rustdoc comment on each of these would be great - especially the `try_` variants to explain why you might want to use one over the other

lib/io/hyped_can/src/lib.rs
```diff
@@ -0,0 +1,97 @@
...
+        pub fn new(
+            frames_to_read: &'static Mutex<CriticalSectionRawMutex, RefCell<CanValues>>,
+        ) -> Self {
+            MockCan {
+                frames_to_read,
+                frames_sent: CanValues::new(),
+            }
+        }
+        /// Get the values that have been sent over the CAN bus
+        pub fn get_can(&self) -> &CanValues {
```
davidbeechey: `get_sent_can_frames`?

lib/io/hyped_can/hyped_can_derive/src/lib.rs
```diff
@@ -0,0 +1,67 @@
...
+use quote::quote;
+use syn::{self, Generics};
+
+#[proc_macro_derive(HypedCan)]
+pub fn hyped_can_derive(input: TokenStream) -> TokenStream {
+    let ast = syn::parse(input).unwrap();
+    impl_hyped_can(&ast)
+}
+
+fn impl_hyped_can(ast: &syn::DeriveInput) -> TokenStream {
```
davidbeechey, line 13: You're missing implementations for `try_read_frame` and `try_write_frame` here

lib/io/hyped_can/hyped_can_derive/src/lib.rs
```diff
@@ -0,0 +1,67 @@
...
+    let name = &ast.ident;
+    let generics: &Generics = &ast.generics;
+    let (impl_generics, ty_generics, _) = generics.split_for_impl();
+    let gen = quote! {
+        impl #impl_generics HypedCan for #name #ty_generics{
+            /// Read a byte from a register on a device
+            fn read_frame(&mut self, device_address: u8, register_address: u8) -> Option<u8> {
```
davidbeechey: remnants from the I2C version here I think...

lib/io/hyped_can/hyped_can_derive/src/lib.rs
```diff
@@ -0,0 +1,67 @@
...
+                let mut read = [0];
+                let result = self.can.lock(|can| {
+                    can.borrow_mut().blocking_write_read(
+                        device_address,
+                        [register_address].as_ref(),
+                        &mut read,
+                    )
```
davidbeechey: likewise here, `self.can` has `read` that doesn' have any arguments and returns `Result<Envelope, BusError>`: https://docs.embassy.dev/embassy-stm32/git/stm32f767zi/can/struct.Can.html#method.read

davidbeechey (changes requested):
Well done, nearly there just a few things here and there to be fixed/added

lib/io/hyped_can/hyped_can_derive/src/lib.rs
jpfbastos: - [x] Get Envelope struct going
- [x] Scrap TransmitStatus, return nothing for normal, return result for try_

boards/stm32f767zi/Cargo.lock
TomLonergan03, line 1: why did the lockfiles get deleted?

lib/io/hyped_can/hyped_can_derive/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+                        embassy_stm32::can::enums::BusError::BitDominant => CanError::BitDominant,
+                        embassy_stm32::can::enums::BusError::Crc => CanError::Crc,
+                        embassy_stm32::can::enums::BusError::Software => CanError::Software,
+                        embassy_stm32::can::enums::BusError::BusOff => CanError::BusOff,
+                        embassy_stm32::can::enums::BusError::BusPassive => CanError::BusPassive,
+                        embassy_stm32::can::enums::BusError::BusWarning => CanError::BusWarning,
+                    }),
+                    Err(embassy_stm32::can::enums::TryReadError::Empty) => Err(CanError::Empty),
+                }
+            }
```
TomLonergan03: one thing you could do here, is `impl From<BusError> for CanError` where `CanError` is defined, then you can just return `e.into()` here. that might be overkill tho

lib/io/hyped_can/hyped_can_derive/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+                            false,
+                        ),
+                        id => embassy_stm32::can::frame::Header::new(
+                            embassy_stm32::can::Id::Extended(
+                                embassy_stm32::can::ExtendedId::new(id).unwrap(),
+                            ),
+                            frame.data.len() as u8,
+                            false,
+                        ),
+                    },
```
TomLonergan03: do this before the frame, this is too much nesting

lib/io/hyped_can/hyped_can_derive/src/lib.rs
```diff
@@ -0,0 +1,105 @@
...
+                    }
+                    Err(e) => Err(match e {
+                        embassy_stm32::can::enums::FrameCreateError::NotEnoughData => {
+                            CanError::NotEnoughData
+                        }
+                        embassy_stm32::can::enums::FrameCreateError::InvalidDataLength => {
+                            CanError::InvalidDataLength
+                        }
+                        embassy_stm32::can::enums::FrameCreateError::InvalidCanId => CanError::InvalidCanId,
+                    }),
```
TomLonergan03: could `impl Into...` for this as well

lib/io/hyped_can/hyped_can_derive/src/lib.rs
TomLonergan03, line 1: change all the huge qualification chains e.g. `embassy_stm32::can::enums::...` into `use`s. Import the enum and use its variants etc.

unless that isn't the standard anywhere else but i think it is

lib/io/hyped_can/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+pub trait HypedCan {
+    /// Attempts to read a CAN frame without blocking.
+    ///
+    /// Returns [Err(TryReadError::Empty)] if there are no frames in the rx queue.
+    fn read_frame(&mut self) -> Result<HypedEnvelope, CanError>;
+    /// Attempts to transmit a frame without blocking.
+    ///
+    /// Returns [Err(CanError::Full)] if the frame can not be queued for transmission now.
+    ///
+    /// If FIFO scheduling is enabled, any empty mailbox will be used.
```
TomLonergan03: are you supporting enabling and disabling fifo scheduling?

otherwise i recognise this from the embassy docs lol

TomLonergan03 (changes requested):

boards/stm32f767zi/Cargo.lock
jpfbastos: weird merge conflicts pretty much (@davidbeechey thanks for handling that and the long imports)

lib/io/hyped_can/hyped_can_derive/src/lib.rs
jpfbastos: yeah i tried doing that - however i needed to add `embassy-stm32` to the Cargo.toml file, which can't be done as it would make it board specific

lib/io/hyped_can/src/lib.rs
jpfbastos: good catch :) i copied the description on embassy, I'll remove that

lib/io/hyped_can/hyped_can_derive/src/lib.rs
TomLonergan03: ah ok, does that also happen if you don't enable a board in the features?

jpfbastos: not sure if i fully understood the question, but basically the moment you write `impl From<CanError> for embassy_stm32::can::BusError` for example, I get `failed to resolve: use of undeclared crate or module embassy_stm32 
use of undeclared crate or module embassy_stm32`

so i would have to import it in that case, but not sure if that's what you were asking

davidbeechey: I can clarify a couple of things here:
- Reason why we can't use `From` here is because the embassy type `BusError` and our type `HypedCanError` are both defined in other crates and rust doesn't let you do that.
- The options then become adding the `From` stuff to the crate where `HypedCanError` is defined, which would require importing `embassy_stm32` (but we can't import `embassy_stm32` without enabling exactly one board feature), or moving the `HypedCanError` enum to the code in `hyped_can_derive` which I don't think is possible because that would create separate versions for each board's IO where the macro is used

jpfbastos:
I'll sort this out later today

davidbeechey:
@jpfbastos this is looking good, but I'd like to test it in the lab before we merge it. Since we need it to move on with IMD, BMS, CanOpen, etc. could you open a new PR just with the `HypedCan` trait (leaving out `hyped_can_derive`)? That way we can get started without needing a completed+tested CAN hardware implementation. Thanks!

jpfbastos:
> @jpfbastos this is looking good, but I'd like to test it in the lab before we merge it. Since we need it to move on with IMD, BMS, CanOpen, etc. could you open a new PR just with the `HypedCan` trait (leaving out `hyped_can_derive`)? That way we can get started without needing a completed+tested CAN hardware implementation. Thanks!

Done - can be found in the "Can IO simple" PR, lmk if there's anything else I need to do!

lib/io/hyped_can/src/lib.rs
```diff
@@ -0,0 +1,149 @@
...
+    BusWarning,
+    Full,
+    Empty,
+    Unknown,
+    NotEnoughData,
+    InvalidDataLength,
+    InvalidCanId,
+}
+
+#[derive(Clone)]
```
TomLonergan03: i think this could derive `Copy` also if you want, avoids needing to use `.clone()` cos it's simple types that can be memcopied

TomLonergan03 (approved):

davidbeechey (approved):

davidbeechey:
Tested in the lab 19/02/25-20/02/25 using the `can_io_recv_test` and `can_io_send_test` binaries :) All working

PR #65, opened 2025-01-27
INFRA - Implement CAN encoding

platinumxy:


lib/core/src/can_sendable.rs
```diff
@@ -0,0 +1,292 @@
...
+
+pub trait CanSendable {
+    fn encode_to_can(&self, board_id: u8) -> [u8; 8];
+    fn can_decode(can_msg: &[u8; 8]) -> Self;
+}
+
+impl CanSendable for bool {
+    fn encode_to_can(&self, board_id: u8) -> [u8; 8] {
+        let mut data: [u8; 8] = [0; 8];
+        data[0] = build_can_header(board_id, CAN_MSG_TYPE_BOOL);
```
TomLonergan03: This should be an enum becuase there's only the specific set values that have meaning.

lib/core/src/can_sendable.rs
```diff
@@ -0,0 +1,292 @@
...
+    }
+
+    /// Check we've received a completed PositionDelta, (all x,y,z all with the same clock value)
+    pub fn is_complete(&self) -> bool {
+        self.clock.is_some() && self.x.is_some() && self.y.is_some() && self.z.is_some()
+    }
+
+    /// Returns finished f32
+    pub fn return_complete(&self) -> [f32; 3] {
+        assert!(
```
TomLonergan03: This should return an optional instead so it doesn't crash at runtime, and you then don't need to call is complete, you can just look at whether you get `Some()` back

lib/core/src/can_sendable.rs
```diff
@@ -0,0 +1,292 @@
...
+            _ => {
+                error!("Unknown PositionDelta step type: {}", step_type);
+                panic!();
+            }
+        }
+        None
+    }
+
+    /// Checks if new clock is infront of the old
+    pub fn is_clock_infuture(old: u8, new: u8) -> bool {
```
TomLonergan03: this doesn't need to be `pub`

TomLonergan03 (changes requested):

TomLonergan03 (approved):

lib/core/src/can_sendable.rs
```diff
@@ -0,0 +1,287 @@
+use defmt::error;
+
+pub enum CanMsgType {
```
davidbeechey:
```suggestion
pub enum CanMessageType {
```

davidbeechey (changes requested):

davidbeechey (approved):

PR #68, opened 2025-01-30
SNS - High Pressure Sensor

siny-over-cosy:


lib/sensors/src/high_pressure.rs
```diff
@@ -0,0 +1,36 @@
...
+
+/// The high pressure sensor (SPAW-P25R-G12M-2N-M12) is able to detect pressure in range from 0 to 25 bar
+/// 
+
+/// Links to datasheets
+///     (https://www.festo.com/media/catalog/203715_documentation.pdf)
+///     (https://ftp.festo.com/public/PNEUMATIC/SOFTWARE_SERVICE/DataSheet/EN_GB/8022773.pdf)
+
+pub struct HighPressure<T: HypedAdc> {
+    pressure: u16,
```
davidbeechey: don't think we need to keep the pressure reading in the struct - just return it

davidbeechey (commented):
Looking good, just need to talk to electronics and/or the pneumatics team in dynamics to work out what the formula should be

lib/sensors/src/high_pressure.rs
```diff
@@ -0,0 +1,56 @@
...
+        sp2_gpio: T,
+    ) -> HighPressure<T> {
+        HighPressure {
+            sp1_gpio,
+            sp2_gpio,
+        }
+    }
+
+    /// Read SP1 and SP2 GPIO pin values and bitwise OR them. Return state of high pressure sensor based on value of OR'd value.
+    pub fn get_high_pressure_state(&mut self) -> Result<State, &'static str> {
```
davidbeechey: to keep this consistent with our other sensors it'd be better to create a `HighPressureError` enum and use it as the return error type

lib/sensors/src/high_pressure.rs
```diff
@@ -0,0 +1,56 @@
...
+        let sp2 = (self.sp2_gpio.is_high() as u8) << 1;
+
+        let pres_state = sp1 | sp2;
+
+        match pres_state {
+            0 => Ok(State::Off),                        // 00 - none are high
+            1 => Ok(State::State2),                     // 01 - sp1 is high
+            3 => Ok(State::State3),                     // 11 - sp1 and sp2 are high
+            _ => Err("ERROR?? Value: {pres_state}."),   // any other case - should never be possible
+        }
```
davidbeechey: I think this could be simplified to something like this:
```rust
match (self.sp1_gpio.is_high(), self.sp2_gpio.is_high()) {
    (false, false) => ...
    (true, false) => ...
    ...
}
```
or similar to avoid converting into `u8` unnecessarily

lib/sensors/src/high_pressure.rs
```diff
@@ -0,0 +1,56 @@
...
+    }
+}
+
+/// Represents the possible state of the high pressure sensor
+pub enum State {
+    Off,
+    State2,
+    State3
```
davidbeechey: we'll need to rename these are some point, maybe for now `LowRange`, `MidRange` and `HighRange`?

davidbeechey (changes requested):

lib/sensors/src/high_pressure.rs
siny-over-cosy: def a lot better, will implement this

boards/stm32f767zi/src/tasks/read_high_pressure.rs
```diff
@@ -0,0 +1,20 @@
...
+    sp1_pin: Input<'static>,
+    sp2_pin: Input<'static>,
+    sender: Sender<'static, CriticalSectionRawMutex, u32, 1>,
+) -> ! {
+    let mut high_pressure_sensor = HIghPressure::new(Stm32f767ziGpioInput::new(sp1_pin), Stm32f767ziGpioInput::new(sp2_pin));
+
+    loop {
+        high_pressure_sensor.get_high_pressure_state();
+        sender.send(keyence.get_stripe_count());
+        Timer::after(Duration::from_millis(100)).await;
```
davidbeechey: This is probably in progress, but you'll need to `match` on the output of `get_high_pressure_state` and then send it using `sender.send()` instead. Also in other tasks we're setting an `UPDATE_FREQUENCY` constant and then sleeping for that duration inside the loop (see [here](https://github.com/Hyp-ed/hyped-2025/blob/main/boards/stm32f767zi/src/tasks/read_temperature.rs#L52) as an example)

davidbeechey (commented):
Looking good! Could you create a `high_pressure_test` binary in `boards/stm32f767zi/bin` too? Thanks!

siny-over-cosy:
no idea what the problem is, will fix some other time

siny-over-cosy:
the integrated tests are taking an unusually longer time to check after including defmt library.
is it anything to be concerned about or am i just being paranoid?
this isn't to do with the 6 checks that don't finish.

boards/stm32f767zi/src/tasks/read_high_pressure.rs
```diff
@@ -0,0 +1,28 @@
...
+    );
+
+    // update frequency of high pressure sensor in hz
+    const UPDATE_FREQUENCY: u64 = 1000;
+
+    loop {
+        match high_pressure_sensor.get_high_pressure_state() {
+            Ok(high_pressure_state) => sender.send(Ok(high_pressure_state)),
+            Err(e) => sender.send(Err(e)),
+        }
```
davidbeechey: We'll want to handle this error rather than just sending it, for now this can just be `panic!("Invalid high pressure state")` or similar. At the moment this match doesn't have any effect, it's the same as:
```rust
sender.send(high_pressure_sensor.get_high_pressure_state())
```

boards/stm32f767zi/src/tasks/read_high_pressure.rs
```diff
@@ -0,0 +1,28 @@
...
+    sp2_pin: Input<'static>,
+    sender: Sender<'static, CriticalSectionRawMutex, Result<State, &'static str>, 1>,
+) -> ! {
+    let mut high_pressure_sensor = HighPressure::new(
+        Stm32f767ziGpioInput::new(sp1_pin),
+        Stm32f767ziGpioInput::new(sp2_pin),
+    );
+
+    // update frequency of high pressure sensor in hz
+    const UPDATE_FREQUENCY: u64 = 1000;
```
davidbeechey: this can be moved outside the task, also if the comment starts with `///` not `//` then it will be a Rustdoc comment and display when you hover over `UPDATE_FREQUENCY` which is a bit nicer

lib/sensors/src/high_pressure.rs
davidbeechey: like here:
https://github.com/Hyp-ed/hyped-2025/blob/main/lib/sensors/src/laser_triangulation.rs#L55

lib/sensors/src/high_pressure.rs
```diff
@@ -0,0 +1,92 @@
...
+    /// Read SP1 and SP2 GPIO pin values and bitwise OR them. Return state of high pressure sensor based on value of OR'd value.
+    pub fn get_high_pressure_state(&mut self) -> Result<State, &'static str> {
+        let sp1 = self.sp1_gpio.is_high();
+        let sp2 = self.sp2_gpio.is_high();
+
+        match (sp1, sp2) {
+            (false, false) => Ok(State::LowRange),
+            (true, false) => Ok(State::MidRange),
+            (true, true) => Ok(State::HighRange),
+            _ => Err("ERROR??"), // any other case - should never be possible
```
davidbeechey: and here you would have `Err(HighPressureError::InvalidState)`

davidbeechey (changes requested):
Nearly there, looks good

siny-over-cosy:
second revision of finished src code, not sure if you just want me to resolve the merge conflicts
not sure what to do for Cargo.toml

TomLonergan03 (approved):
How are the thresholds for the low/medium/high signals set? Is that electronics or is there a configuration process

davidbeechey (approved):

miksquared (approved):

siny-over-cosy:
> How are the thresholds for the low/medium/high signals set? Is that electronics or is there a configuration process

Apologies for the late response, I couldn't figure out how to add a reply to this.
This will be handled by Electronics. The high pressure sensor has 2 input pins that only read HIGH/LOW. The actual pressure threshold for when inputs are HIGH or LOW will be handled by the Electronics team.

Aux1r (approved):
Had a scroll through today, overall LGTM :)

H-Allen (approved):

PR #69, opened 2025-01-30
SNS - IMD

miksquared:
For SIM100MOD.

Data sheet: https://www.sensata.com/sites/default/files/a/sensata-el-sim100-mod-datasheet.pdf


lib/sensors/src/imd.rs
```diff
@@ -0,0 +1,48 @@
...
+            can,
+            resistance_positive: 0,
+            resistance_negative: 0,
+            isolation_status: 0,
+        } // may need to do error handling
+    }
+
+    pub fn update_values(&mut self) -> Result<(), ImdError> {
+        let frame = HypedCanFrame {
+            can_id: 0x80000000 | IMD_REQUEST_DATA_CAN_ID, //the 0x80000000 is the CAN_EFF_FLAG constant from the 2023 code.
```
davidbeechey: any reason why this isn't as a constant here?

lib/sensors/src/imd.rs
```diff
@@ -0,0 +1,48 @@
+use hyped_can::{CanError, HypedCan, HypedCanFrame};
```
davidbeechey:
```suggestion
use hyped_can::{CanError, HypedCan, HypedCanFrame};

```

lib/sensors/src/imd.rs
```diff
@@ -0,0 +1,48 @@
...
+
+impl<'a, T: HypedCan> Imd<'a, T> {
+    /// Create a new instance of the IMD.
+    pub fn new(can: &'a mut T) -> Self {
+        Self {
+            can,
+            resistance_positive: 0,
+            resistance_negative: 0,
+            isolation_status: 0,
+        } // may need to do error handling
```
davidbeechey: Shouldn't need any error handling here

lib/sensors/src/imd.rs
davidbeechey, line 1: you need to add the 3 "getters" from the 2023 code

davidbeechey (changes requested):
This looks really good, can't seem much that is needed compared to 2023 version

lib/sensors/src/imd.rs
miksquared: i might have just forgot tbh

TomLonergan03 (approved):

davidbeechey (approved):
Looks good I think, just needs testing now

boards/stm32f767zi/src/tasks/read_imd.rs
```diff
@@ -0,0 +1,82 @@
...
+                    imd.get_isolation_status(),
+                    imd.get_resistance_positive(),
+                    imd.get_resistance_negative()
+                );
+            }
+            Err(_) => {
+                println!("Failed to receive frame from IMD");
+            }
+        }
+    }
```
davidbeechey, line 81: I'm not 100% sure how we're going to run this (i.e. whether we're constantly monitoring or requesting updates periodically), but it would feel nicer to loop to wait until there is a value available rather than wait for an arbitrary amount of time

davidbeechey (changes requested):
Looks good - I will bring the IMD to Monday's meeting so you can try to test it!

davidbeechey:
Unfortunately this IMD is not going to be used this year because it doesn't comply with the EHW R&R :(

PR #74, opened 2025-02-09
TEL - Central config

arjunnaha:
Implements new central config schema.

Main schema changes

- `label` instead of `name` to make it clear this is a human-readable label
- `id` instead of a mix of `id` and `key`
- new `statuses` property instead of jamming them into `measurements`

On the telemetry side, this will probably break a lot of server/frontend code, but we will follow this PR up with fixes.


davidbeechey (approved):
good luck fixing everything :saluting_face:

samad-a (approved):

PR #76, opened 2025-02-09
SNS - Laser Triangulation Sensor

Aux1r:
Implements logic for reading the **analogue output** from the Althen laser triangulation sensor: https://www.althensensors.com/sensors/laser-position-sensors/single-point-laser-distance-sensors/fdrf602-laser-triangulation-sensor/.

Data sheet is not available online, but the PDF will be in Slack / Google Drive somewhere...

In the future we might also support the MODBUS digital output from the sensor.


lib/sensors/src/laser_triangulation.rs
```diff
@@ -0,0 +1,47 @@
...
+    pub fn read(&mut self) -> SensorValueRange<f32> {
+        let current = self.adc.read_value() as f32;
+        let result = ((current - AMP_MIN) / (AMP_MAX - AMP_MIN)) * (MEASURE_RANGE - BASE_DISTANCE)
+            + BASE_DISTANCE;
+        (self.calculate_bounds)(result)
+    }
+}
+
+pub fn default_calculate_bounds(value: f32) -> SensorValueRange<f32> {
+    if value <= BASE_DISTANCE || value >= MEASURE_RANGE {
```
davidbeechey: I think you're upper bound will need to be `BASE_DISTANCE + MEASURE_RANGE`?

boards/stm32l432kc/src/tasks/laser_triangulation.rs
davidbeechey, line 1: tasks are usually named `read_laser_triangulation`

boards/stm32l432kc/src/bin/laser_triangulation_test.rs
```diff
@@ -0,0 +1,52 @@
...
+static LASER_TRIANG_READING: Watch<CriticalSectionRawMutex, SensorValueRange<f32>, 1> =
+    Watch::new();
+
+#[embassy_executor::main]
+async fn main(spawner: Spawner) -> {
+    let p = embassy_stm32::init(Default::default());
+    let adc = Adc::new(p.ADC1, Delay);
+
+    // Create a sender to pass to the laser triangulation sensor reading task, and a receiver for reading the values back.
+    let laser_triang_reading_sender = LASER_TRIANG_READING.sender();
```
davidbeechey:
```suggestion
    let laser_triangulation_reading_sender = LASER_TRIANG_READING.sender();
```

davidbeechey: or `laser_distance` or similar, I'm just not a fan of abbreviations (picky)

davidbeechey (changes requested):

lib/sensors/src/laser_triangulation.rs
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    /// The Laser Triangulation sensor has multiple configurations; we're using one where
+    /// the base distance is 20mm, measuring range is 25mm. The output amperage is between 4mA and 20mA,
+    /// and assuming linearity, the function reads in the amperage and maps it to an appropriate range.
+    /// If we change configuration (e.g. different base distance & measuring distance), you'll have to change
+    /// the MEASURE_RANGE and BASE_DISTANCE values accordingly. You can find different configuration settings
+    /// in the data sheet.
+    pub fn read(&mut self) -> SensorValueRange<f32> {
+        let current = self.adc.read_value() as f32;
```
davidbeechey: need to convert to current also here

davidbeechey (changes requested):

boards/stm32l432kc/src/bin/laser_triangulation_test.rs
```diff
@@ -0,0 +1,52 @@
...
+
+    // Every 100ms we read for the latest value from the laser triangulation sensor.
+    loop {
+        match laser_triangulation_reading_receiver.try_changed() {
+            Some(reading) => match reading {
+                Safe(value) => {
+                    defmt::info!("Range: {} mm (safe)", value)
+                }
+                Warning(value) => {
+                    defmt::info!("Range: {} mm (safe)", value)
```
TomLonergan03: should this be safe or warning?

lib/sensors/src/laser_triangulation.rs
```diff
@@ -0,0 +1,48 @@
...
+        let result = ((current - AMP_MIN) / (AMP_MAX - AMP_MIN))
+            * ((MEASURE_RANGE + BASE_DISTANCE) - BASE_DISTANCE)
```
TomLonergan03: is this right? `(MEASURE_RANGE + BASE_DISTANCE) - BASE_DISTANCE` is equal to `MEASURE_RANGE`

lib/sensors/src/laser_triangulation.rs
```diff
@@ -0,0 +1,48 @@
...
+    adc: T,
+    calculate_bounds: fn(f32) -> SensorValueRange<f32>,
+}
+
+impl<T: HypedAdc> LaserTriangulation<T> {
+    /// Create a new instance of the Laser Triangulation sensor
+    pub fn new(adc: T) -> LaserTriangulation<T> {
+        LaserTriangulation {
+            adc,
+            calculate_bounds: default_calculate_bounds,
```
TomLonergan03: there isn't a way to construct this and pass in non-default bounds, so you might need a `new_with_bounds(...)` fn or just inline the bounds calculation (I think you want the first option if i understand what is going on)

TomLonergan03 (changes requested):

boards/stm32l432kc/src/bin/laser_triangulation_test.rs
Aux1r: safe, david said that there isn't really a warning value for the distance (but if you want to suggest one, feel free!)

lib/sensors/src/laser_triangulation.rs
Aux1r: Yea, for some reason I wanted to make it explicitly the same as the formula for mapping values `Y = (X-A)/(B-A) * (D-C) + C`, and our upper bound (`D`) is the `MEASURE_RANGE + BASE_DISTANCE`. Will fix!

boards/stm32l432kc/src/bin/laser_triangulation_test.rs
davidbeechey: it should be `(warning)` for the warning case - even though we aren't using it now, if we do add a warning bound this print will be wrong

lib/sensors/src/laser_triangulation.rs
davidbeechey: agreed - having a `new_with_bounds` is what we did for temperature

TomLonergan03 (approved):

lib/sensors/src/laser_triangulation.rs
```diff
@@ -33,8 +33,11 @@ impl<T: HypedAdc> LaserTriangulation<T> {
     /// the MEASURE_RANGE and BASE_DISTANCE values accordingly. You can find different configuration settings
     /// in the data sheet.
     pub fn read(&mut self) -> SensorValueRange<f32> {
-        let current = self.adc.read_value() as f32;
-        let result = ((current - AMP_MIN) / (AMP_MAX - AMP_MIN)) * (MEASURE_RANGE) + BASE_DISTANCE;
+        let adc_reading = self.adc.read_value() as f32;
+        let resolution = self.adc.get_resolution() as f32;
+        let v_ref = self.adc.get_reference_voltage();
+        let voltage = (adc_reading / resolution) * v_ref;
```
TomLonergan03: is there a reason to do this outside of the adc? all the values are coming from the adc

lib/sensors/src/laser_triangulation.rs
```diff
@@ -33,8 +33,11 @@ impl<T: HypedAdc> LaserTriangulation<T> {
...
+        let result = (voltage / v_ref) * (MEASURE_RANGE + BASE_DISTANCE);
```
TomLonergan03: is this correct? i feel like it should be 
```rs
let result = (voltage / v_ref) * MEASURE_RANGE + BASE_DISTANCE;
```

TomLonergan03 (changes requested):
apologies for requesting more changes after approving a while back

lib/sensors/src/laser_triangulation.rs
Aux1r: Good spot I think you're right

lib/sensors/src/laser_triangulation.rs
Aux1r: Good point, I'll add a get_voltage function for adc that'll do this instead of the 4-liner

TomLonergan03 (approved):

davidbeechey (approved):

PR #77, opened 2025-02-09
CTL - Updated Levitation Control

mostafam99:
Keep as Draft PR until more info from Levitation


boards/stm32l432kc/src/bin/levitation.rs
```diff
@@ -31,6 +32,67 @@ const GAIN_CURRENT: PidGain = PidGain {
...
+    kp: 1.1,
+    ki: 0.12,
+    kd: 0.05,
+};
+
+/// `Pid` is a structure that implements the [`PidController`] trait.
+#[derive(Debug, Clone)]
+pub struct Pid {
+    config: PidConfig,
+    i_term: FloatType,
```
jpfbastos: is it possible to use `f32` instead of `FloatType`?

boards/stm32l432kc/src/bin/levitation.rs
```diff
@@ -31,6 +32,67 @@ const GAIN_CURRENT: PidGain = PidGain {
...
+    previous_filter: FloatType,
+}
+
+impl PidController for Pid {
+    /// Creates a new `Pid` with the specified configuration.
+    fn new(config: PidConfig) -> Self {
+        Self {
+            config,
+            i_term: 0.0,
+            pre_error: FloatType::NAN,
```
jpfbastos: for the `NAN` you can use f32::NAN

boards/stm32l432kc/src/bin/levitation.rs
```diff
@@ -31,6 +32,67 @@ const GAIN_CURRENT: PidGain = PidGain {
...
+            current_filter: 0.0,
+            previous_filter: 0.0,
+        }
+    }
+    /// Updates the `Pid` controller with the specified set point, actual value, and time delta.
+    /// Implements a low pass filter onto the derivative term.
+    /// Returns the controller output.
+    fn update_wfilter(&mut self, set_point: FloatType, actual: FloatType, dt: FloatType, filter_constant: FloatType) -> FloatType {
+        let error = set_point - actual;
```
jpfbastos: here, you'll probably have to do 3 sets of errors, as in the whiteboard you can see that the reference value (i.e. `set_point`) has a constant you multiply it by for P and D

maybe consider adding these two constants to the PID struct

boards/stm32l432kc/src/bin/levitation.rs
```diff
@@ -45,6 +107,7 @@ then performed, and the duty cycle is set, and timer restarted.
 async fn main(_spawner: Spawner) {
     let mut pid_height = Pid::new(GAIN_HEIGHT.into());
     let mut pid_current = Pid::new(GAIN_CURRENT.into());
+    let mut pid_voltage = Pid::new(GAIN_VOLTAGE.into());
```
jpfbastos: rename these `pi_current` and `pi_voltage` as it is easier to understand their types

jpfbastos (changes requested):
Great! Only these minor fixes, but otherwise you're good

jpfbastos:
Cool - we'll show lev this next week to see what they think

davidbeechey (commented):
Great job, looks promising - hopefully it's what levitation are looking for :)

davidbeechey (changes requested):
Split out the PID logic into a new PID crate:
- simplifies the code in this file
- makes it easier for you to unit test the PID logic

boards/stm32l432kc/src/bin/levitation.rs
```diff
@@ -1,8 +1,9 @@
 #![no_std]
 #![no_main]
 
-use advanced_pid::{prelude::*, Pid, PidGain};
+use core::f32;
```
davidbeechey: is this necessary?

lib/control/src/pid_controller.rs
```diff
@@ -0,0 +1,133 @@
+use core::f32;
```
davidbeechey: same here?

davidbeechey (approved):

lib/control/src/pid_controller.rs
mostafam99: JP here - Basically it was there because we use `f32::NAN` for initialising the `pre_error`. I thought the normal version was using standard library, so I was using the core version. However, both the `core` and `std` versions are deprecated, and they're being just stored in the f32 primitive, so it can be scrapped and we can use `f32::NAN` by itself

jpfbastos (approved):

davidbeechey (approved):
I'm happy with this as long as it's been looked over by levitation. I don't mind if you want to merge this for now even if it's subject to change later on or keep it open until it's finalised. If you want to keep it open for the levitation control stuff the new PID code could be put in a different PR? Again up to you

boards/stm32l432kc/src/bin/levitation.rs
```diff
@@ -10,40 +9,50 @@ use embassy_stm32::timer::low_level::CountingMode;
...
-    ki: 0.12,
-    kd: 0.05,
+const GAIN_CURRENT: PiGain = PiGain {
+    kp: 1000.0,
+    ki: 5_000_000.0,
+};
+
+const GAIN_VOLTAGE: PiGain = PiGain {
+    kp: 50.0,
+    ki: 10_000_000.0,
```
TomLonergan03: is this stuff for the config or is that only for things that need tel interop?

boards/stm32l432kc/src/bin/levitation.rs
```diff
@@ -70,21 +79,27 @@ async fn main(_spawner: Spawner) {
...
+            (pi_current.update(target_current, actual_current, dt)).min(MAX_VOLTAGE); // takes in current -> outputs voltage (within boundaries) and ignores derivative term from output
 
-        let required_voltage =
-            (pid_current.update(target_current, actual_current, dt)).min(MAX_VOLTAGE); // takes in current -> outputs voltage (within boundaries)
+        let duty_cycle = pi_voltage.update(target_voltage, actual_voltage, dt); // TODOLater include .min(max_duty) if max_duty given
 
-        let duty_cycle = max_duty * (required_voltage / MAX_VOLTAGE); // the duty cycle ranges from 0 to max_duty, so what fraction of that do we need
-                                                                      // probably TODOLater update how this is calculated
+        let duty_cycle = duty_cycle * max_duty; // the duty cycle ranges from 0 to max_duty, so what fraction of that do we need
+                                                // probably TODOLater update how this is calculated
```
TomLonergan03: why do you have the todolater on `.min(max_duty)` along with multiplying by max duty? and what is wrong with how it's calculated currently?

lib/control/src/pid_controller.rs
```diff
@@ -0,0 +1,71 @@
+pub trait PidController {
```
TomLonergan03: why two different traits? is there a need for the trait at all?

lib/control/src/pid_controller.rs
```diff
@@ -0,0 +1,71 @@
...
+    fn new(config: PidGain) -> Self;
```
TomLonergan03: i wouldn't put constructors in a trait, cos if you implement it again you will probably want to pass in other values for the constructor.

e.g. if you have a `HardcodedPidController` for testing something you might want to pass in a list of preset values for it to emit

also, the only place you would use a constructor that's behind a trait is if you have some nasty generic thing that is prob a sign to rethink what you're doing anyway

lib/control/src/pid_controller.rs
```diff
@@ -0,0 +1,71 @@
...
+    fn update(&mut self, set_point: f32, actual: f32, dt: f32, filter_constant: f32) -> f32;
```
TomLonergan03: i don't think the filter constant should be passed in on each iteration, instead it should be passed to the constructor, unless the constant is changing each iter (if the constant is changing rename it lol)

if you don't have this you could have one `Controller` trait that is shared by both should you wish

lib/control/src/pid_controller.rs
```diff
@@ -0,0 +1,71 @@
...
+    pub ki: f32,
+    pub kd: f32,
+    pub p_reference_gain: f32,
+    pub d_reference_gain: f32,
+}
+/// `Pid` is a structure that implements the [`PidController`] trait.
+#[derive(Debug, Clone)]
+pub struct Pid {
+    pub config: PidGain,
+    pub i_term: f32,
```
TomLonergan03: what is `i_term`? if i remember pid right it's better as `integral_term`

lib/control/src/pid_controller.rs
```diff
@@ -0,0 +1,71 @@
...
+
+#[derive(Debug, Clone)]
+pub struct PidGain {
+    pub kp: f32,
+    pub ki: f32,
+    pub kd: f32,
+    pub p_reference_gain: f32,
+    pub d_reference_gain: f32,
+}
+/// `Pid` is a structure that implements the [`PidController`] trait.
```
TomLonergan03: unnecessary comment

lib/control/src/pid_controller.rs
```diff
@@ -0,0 +1,71 @@
...
+#[derive(Debug, Clone)]
+pub struct Pid {
+    pub config: PidGain,
+    pub i_term: f32,
+    pub pre_error: f32,
```
TomLonergan03: what is `pre_error`?

lib/control/src/pid_controller.rs
```diff
@@ -0,0 +1,71 @@
...
+    pub current_filter: f32,
+}
+
+impl PidController for Pid {
+    /// Updates the `Pid` controller with the specified set point, actual value, and time delta.
+    /// Implements a low pass filter onto the derivative term.
+    /// Returns the controller output.
```
TomLonergan03: is this meant to be a comment for `update(...)`?

lib/control/src/pi_controller.rs
```diff
@@ -0,0 +1,53 @@
...
+            pre_error: f32::NAN,
+        }
+    }
+    /// Updates the `Pi` controller, ignoring D.
+    /// Returns the controller output.
+    fn update(&mut self, set_point: f32, actual: f32, dt: f32) -> f32 {
+        let error = set_point - actual;
+        self.i_term += error * dt;
+        let output = self.config.kp * error + self.config.ki * self.i_term; // removed the derivative term
+        self.pre_error = error;
```
TomLonergan03: what is this for? seems to be set and never read

lib/control/src/pi_controller.rs
```diff
@@ -0,0 +1,53 @@
...
+pub struct PiGain {
+    pub kp: f32,
+    pub ki: f32,
+}
+
+#[derive(Debug, Clone)]
+pub struct Pi {
+    pub config: PiGain,
+    pub i_term: f32,
+    pub pre_error: f32,
```
TomLonergan03: think these fields should be private

lib/control/src/pid_controller.rs
```diff
@@ -0,0 +1,71 @@
...
+    pub p_reference_gain: f32,
+    pub d_reference_gain: f32,
+}
+/// `Pid` is a structure that implements the [`PidController`] trait.
+#[derive(Debug, Clone)]
+pub struct Pid {
+    pub config: PidGain,
+    pub i_term: f32,
+    pub pre_error: f32,
+    pub current_filter: f32,
```
TomLonergan03: i think most of these fields don't need `pub` too

TomLonergan03 (changes requested):
few changes, some of it might be me not remembering pid stuff

boards/stm32l432kc/src/bin/levitation.rs
jpfbastos: yep the `.min(max_duty)` I can implement now, and I don't feel there's any problem with the bottom one (i think it's still there from the previous version and i forgot to remove it)

lib/control/src/pid_controller.rs
jpfbastos: it's a filtered derivative thing - basically we're not just doing a normal PID, lev asked us to implement it differently (actually the filtered derivative function changed, so this will change again soon)

lib/control/src/pid_controller.rs
jpfbastos: yes

lib/control/src/lev_controllers.rs
```diff
@@ -0,0 +1,145 @@
+pub trait ControllerTrait {
```
davidbeechey: Don't like having `Trait` in the trait name - just `Controller` would do. I don't think it's strictly necessary to have `PiController` and `PidController` implement a shared trait, but happy to go with it

davidbeechey (changes requested):
Looks good! Only thing left to do is the config

TomLonergan03 (approved):
looks good, work out what's going in the config

jpfbastos:
I feel this is the right idea, but lmk if there's anything to change regarding configs (also I need to look at the trait thing specified above).

Also - I get a "embassy_executor::main: failed to build proc-macro" error in VSCode but it passes all checks and it builds on my laptop, so it may just be my vsc piping up for some reason

config/levitation.yaml
```diff
@@ -0,0 +1,20 @@
+max_voltage_v: 500 # TODOLater these will all change
+max_current_a: 5.0
+target_height_mm: 15.0
+sampling_period: 1 # (decide units) TODOLater see how long this takes to run and go based on that
+
+gain_height:
```
davidbeechey: `gain_height_pid` or `gain_height_pid_constants` or something similar might be more descriptive about what this represents

lib/control/src/lev_controllers.rs
davidbeechey, line 1: I'm not very keen on `lev_controllers.rs` because it's not levitation specific (just generic pi/pid controllers). I also don't like `controllers.rs` because that's too broad, and we can't call it `pid_controllers` because there's also a PI controller in there. My suggestion, apart from creating a new crate just for the pi/pid controllers, is to create a new _module_ inside this `hyped_control` crate called `controllers`. In it you can have a `lib.rs` that defines the `Controller` trait and two _separate_ files for `pi_controller` and `pid_controller`. Let me know what you think?

davidbeechey (changes requested):
Config stuff looks good! Just a few other things to think about

lib/control/src/lev_controllers.rs
jpfbastos: Yes, that's fair - the PI controller is the generic one, but the PID has the filtered derivative element which makes it behave differently to a regular PID. Would it be worth renaming `pid_controller` to something else, or just keep it the way you suggested is and write a comment explaining this?

davidbeechey: up to you, if you can think of a name that's not horrendously long then go for it but a comment should be okay :)

lib/control/src/controllers/mod.rs
```diff
@@ -0,0 +1,11 @@
+pub mod pi_controller;
+pub mod pid_controller;
+
+pub use pi_controller::{PiController, PiGain};
+pub use pid_controller::{PidController, PidGain};
+
+pub use crate::config;
+
+pub trait ControllerTrait {
```
davidbeechey:
```suggestion
pub trait Controller {
```

lib/control/src/lib.rs
```diff
@@ -1,4 +1,8 @@
 #![cfg_attr(not(test), no_std)]
 
+pub mod config;
+pub mod controllers;
 pub mod hp_relay;
 pub mod pneumatics;
+
+pub use controllers::{ControllerTrait, PiController, PiGain, PidController, PidGain};
```
davidbeechey: feels better to just `pub use controllers;` and `use crate::controllers::{PiController}` etc.

davidbeechey (changes requested):
very small, sorry

jpfbastos:
> very small, sorry

yeah no worries - should be done now

davidbeechey (approved):

PR #79, opened 2025-02-09
TEL - Delete unneeded packages

arjunnaha:


davidbeechey (changes requested):
`telemetry-ui` is failing - might be good to fix now?

PR #80, opened 2025-02-09
INFRA - Cross-platform Build Script

platinumxy:


platinumxy:
We need to test on windows, no one with windows env setup was present, will test later

davidbeechey (approved):
I kinda hate it, but if it works it works + it's a bit tidier than 2 files

TomLonergan03 (approved):
lmao

platinumxy:
Why did i call that commit both boards working? I ment to put OSs

build.ps1
```diff
@@ -0,0 +1,48 @@
+echo --% >/dev/null;: ' | out-null
+<#'
+echo "Building main workspace"
```
TomLonergan03, line 3: it slighly bothers me that these print different things on the two OSes lol

TomLonergan03 (commented):
are we changing to use this script in the CI job or keep the matrix there?

davidbeechey:
> are we changing to use this script in the CI job or keep the matrix there?

Keeping the CI job the same - easier to see what jobs are running/failing and doing them in parallel. This script is basically just for “does everything work” during development without having to cd everywhere all the time lol

build.ps1
platinumxy: Ops built both without looking at one another, ill fix that later today

davidbeechey (changes requested):
I'm coming back to this and I'd be happier if this was two build scripts tbh, one for Windows one for Linux/Mac. The two OS in one file thing is cool though, just not right here

PR #82, opened 2025-02-10
SNS - ADC Report Resolution

siny-over-cosy:
Implementing function to get resolution of ADC device

### DIFF
--- a/lib/io/hyped_adc/hyped_adc_derive/src/lib.rs
+++ b/lib/io/hyped_adc/hyped_adc_derive/src/lib.rs
@@ -16,14 +16,13 @@ fn impl_hyped_adc(ast: &syn::DeriveInput) -> TokenStream {
     let (impl_generics, ty_generics, _) = generics.split_for_impl();
     let gen = quote! {
         impl #impl_generics HypedAdc for #name #ty_generics {
-            /// Read a value from the ADC channel
             fn read_value(&mut self) -> u16 {
                 self.adc.blocking_read(&mut self.channel)
             }
 
-            /// Return value of ADC resolution (4095)
             fn get_resolution(&self) -> u16 {
-                4095
+                /// STM32 boards have a resolution of 12 bits
+                4096
             }
         }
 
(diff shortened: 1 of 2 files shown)


lib/io/hyped_adc/src/lib.rs
```diff
@@ -3,6 +3,7 @@
 /// ADC trait used to abstract the ADC peripheral
 pub trait HypedAdc {
     fn read_value(&mut self) -> u16;
+    fn get_resolution(&self) -> u16;
```
davidbeechey: not your fault, but we're moving to having comments for functions here rather than in the derive macro so could you add comments for `Read a value from the ADC channel` and `Get the resolution of the ADC sensor`. Would be good to add something like "this is the max value that will be returned, i.e. 12-bit resolution will return 4095" to make it really clear

davidbeechey: (and remove the comments from the derive stuff)

lib/io/hyped_adc/hyped_adc_derive/src/lib.rs
```diff
@@ -20,6 +20,11 @@ fn impl_hyped_adc(ast: &syn::DeriveInput) -> TokenStream {
             fn read_value(&mut self) -> u16 {
                 self.adc.blocking_read(&mut self.channel)
             }
+
+            /// Return value of ADC resolution (4095)
+            fn get_resolution(&self) -> u16 {
+                4095
```
davidbeechey: since these comments are being removed, would be good to add a comment above `4095` to say "STM32 boards have a resolution of 12-bits" or similar

davidbeechey (changes requested):

lib/io/hyped_adc/hyped_adc_derive/src/lib.rs
davidbeechey: I'm dumb, this should be `4096` everywhere not `4095`

davidbeechey (changes requested):

TomLonergan03 (approved):

lib/io/hyped_adc/src/lib.rs
```diff
@@ -2,7 +2,11 @@
 
 /// ADC trait used to abstract the ADC peripheral
 pub trait HypedAdc {
+    /// Read value from the ADC channel
     fn read_value(&mut self) -> u16;
+
+    /// Get resolution of ADC sensor
```
davidbeechey: sorry this is really picky
```suggestion
    /// Get resolution of ADC
```

davidbeechey (changes requested):

davidbeechey (approved):

PR #83, opened 2025-02-13
CTL - Motor Controller Calculators

jpfbastos:


lib/motors/src/time_frequency_calculator.rs
```diff
@@ -0,0 +1,38 @@
...
+
+/// Takes in a frequency table and calculates the frequency based on the time elapsed since the calculator was created
+/// The frequency table is a list of pairs, where the first element is the time elapsed in microseconds,
+/// and the second element is the frequency to return when the time elapsed is less than the first element.
+///
+/// Useful for creating a frequency that predictably changes over time.
+///
+/// Returns the frequency corresponding to the time elapsed since the calculator was created.
+pub struct TimeFrequencyCalculator {
+    frequency_table: Vec<Vec<u32, 2>, 256>,
```
TomLonergan03: `Vec<(u32, u32), 256>` imo

lib/motors/src/time_frequency_calculator.rs
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    pub fn calculate_frequency(&self) -> u32 {
+        let microseconds_elapsed = Instant::now().as_micros() as u32 - self.start_time;
+        self.frequency_table
+            .iter()
+            .rev()
+            .find(|pair| pair[0] < microseconds_elapsed)
+            .map(|pair| pair[1])
+            .unwrap_or(0)
```
TomLonergan03: 🦀🦀🦀 lfg

lib/motors/src/time_frequency_calculator.rs
```diff
@@ -0,0 +1,38 @@
...
+            frequency_table,
+        }
+    }
+
+    pub fn calculate_frequency(&self) -> u32 {
+        let microseconds_elapsed = Instant::now().as_micros() as u32 - self.start_time;
+        self.frequency_table
+            .iter()
+            .rev()
+            .find(|pair| pair[0] < microseconds_elapsed)
```
TomLonergan03: if you make it vec of `(u32,u32)` then this becomes `|(time, _)| time < microseconds_elapsed` or similar which is nicer

lib/motors/src/time_frequency_calculator.rs
```diff
@@ -0,0 +1,38 @@
...
+            .map(|pair| pair[1])
```
TomLonergan03: similar for this then

lib/motors/src/velocity_frequency_calculator.rs
```diff
@@ -0,0 +1,36 @@
...
+    Overflow(f32),
+}
+
+impl VelocityFrequencyCalculator {
+    pub fn new(coefficients: [f32; 5]) -> Self {
+        VelocityFrequencyCalculator { coefficients }
+    }
+
+    pub fn calculate_frequency(&self, velocity: f32) -> Result<u32, FrequencyError> {
+        let frequency = velocity * velocity * velocity * velocity * self.coefficients[0]
```
TomLonergan03: `velocity.pow(5)`

TomLonergan03 (changes requested):
if you're doing the same approach i did with motors way back (and you defo don't need to, consider other options whether you go with them or not), you'll need all the frequency calculators to share a trait.

probably something like:
```rs
pub trait FrequencyCalculator {
    pub fn calculate_frequency(velocity: f32) -> u8;
}
```
this does mean some of the calculators have to take a irrelevant parameter but for test classes idm that too much

TomLonergan03 (approved):

davidbeechey (approved):

PR #89, opened 2025-02-21
SNS - Keyence Configurable Update Edge

davidbeechey:
Adds a new argument to `Keyence::new()` called `update_on` which represents which digital signal (high or low) the stripe count should be updated on.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -51,15 +51,4 @@ jobs:
           else
             echo "All good"
           fi
-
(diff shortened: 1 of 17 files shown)


TomLonergan03 (approved):
when are you going to want it the other way round?

davidbeechey:
> when are you going to want it the other way round?

Samuel said the way the PCB has been designed means it might be the other way round for some reason (honestly not sure)

H-Allen (approved):

PR #90, opened 2025-02-21
SNS - Temperature Sensor Mux Board

davidbeechey:
Adds a task for reading 24 temperature sensors from the temperature sensor mux board.


davidbeechey:
Progress blocked at the moment because it looks like there's an issue with the PCB (not showing up on `i2cdetect`)

PR #92, opened 2025-02-25
SNS - Split CAN IO into Rx and Tx

davidbeechey:
Splits `HypedCan` into `HypedCanRx` and `HypedCanTx` so that we can have a sensor (e.g. the IMD) just take in a CAN sender so that a separate task can be constantly checking for new CAN messages using a receiver to run a callback.


miksquared (approved):
lgtm

Aux1r (approved):
Looks good to me, only thing that stands out is the `pub trait HypedCanRx` comment about not being used; once we know this should be removed unless I'm missing something.

H-Allen (approved):

PR #93, opened 2025-02-27
SNS - Fix `Watch` issue (set sensor update frequencies)

davidbeechey:
Fixes the issue that @Aux1r found when testing a temperature sensor that the current implementation using `Watch` doesn't do anything.

I believe the issue was that the main loop was never being run because the spawned `read_temperature` task never waited inside its loop, resulting in the main loop never being scheduled (or something to that effect). Resolved by adding an `UPDATE_FREQUENCY` to `read_*` tasks to throttle the loop.

### DIFF
--- a/boards/stm32f767zi/src/tasks/read_temperature.rs
+++ b/boards/stm32f767zi/src/tasks/read_temperature.rs
@@ -9,10 +9,14 @@ use embassy_sync::{
     },
     watch::Sender,
 };
+use embassy_time::{Duration, Timer};
 use hyped_sensors::temperature::{Status, Temperature, TemperatureAddresses};
 use hyped_sensors::SensorValueRange;
 
 type I2c1Bus = Mutex<NoopRawMutex, RefCell<I2c<'static, Blocking>>>;
+
+/// The update frequency of the temperature sensor in Hz
+const UPDATE_FREQUENCY: u64 = 1000;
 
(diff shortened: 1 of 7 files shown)


boards/stm32f767zi/src/tasks/read_temperature.rs
```diff
@@ -9,11 +9,15 @@ use embassy_sync::{
...
     watch::Sender,
 };
+use embassy_time::{Duration, Timer};
 use hyped_sensors::temperature::{Status, Temperature, TemperatureAddresses};
 use hyped_sensors::SensorValueRange;
 
 type I2c1Bus = Mutex<NoopRawMutex, RefCell<I2c<'static, Blocking>>>;
 
+/// The update frequency of the temperature sensor in Hz
+const UPDATE_FREQUENCY: u64 = 1000;
```
TomLonergan03, line 19: possibly a config value? i suspect the jury would nitpick update frequencies somewhere lol

TomLonergan03 (approved):

Aux1r:
looks good!

Aux1r (approved):

boards/stm32f767zi/src/tasks/read_temperature.rs
davidbeechey: Good point - I think I'll leave it for now and deal with all the config stuff later on

PR #94, opened 2025-03-01
INFRA - Inter-board Communications

davidbeechey:
Adds logic for communicating messages between STM32 boards. CAN messages are abstracted. `emergency!()` macro is born. Heartbeats between boards.

Also implements `state_machine` and `state_updater` tasks which respond to transition requests and transitions boards respectively.

To do:
- [x] Test heartbeats between CAN boards
- [x] Test emergency state immediate update on all boards
- ~~How to handle state preconditions~~ will probs do in another PR
- ~~Custom panic handler to trigger emergency?~~
- [x] CAN to MQTT bridge

Issues closed:
- closes HYPE-128 (update state machine states)
- closes HYPE-120 (CAN heartbeats to every board)


boards/stm32f767zi/src/tasks/mqtt/send.rs
```diff
@@ -1,26 +1,28 @@
...
 
 /// Channel for sending messages over MQTT.
 /// Any message sent to this channel will be sent to the MQTT broker by the `mqtt_send_task`
-pub static SEND_TO_MQTT_CHANNEL: Channel<ThreadModeRawMutex, MqttMessage, 128> = Channel::new();
+pub static MQTT_SEND: Channel<ThreadModeRawMutex, MqttMessage, 128> = Channel::new();
 
 /// Task for sending messages from `SEND_CHANNEL` to the MQTT broker
-#[embassy_executor::task]
-pub async fn mqtt_send_task(stack: &'static Stack<Ethernet<'static, ETH, GenericSMI>>) {
+pub async fn mqtt_send(stack: &'static Stack<Ethernet<'static, ETH, GenericSMI>>) {
```
TomLonergan03, line 25: doc is wrong, this isn't a task anymore. or should it be?

lib/communications/src/identifier.rs
```diff
@@ -0,0 +1,34 @@
...
+}
+
+impl From<MessageIdentifier> for u16 {
+    fn from(val: MessageIdentifier) -> Self {
+        match val {
+            MessageIdentifier::Measurement(measurement_id) => measurement_id.into(),
+            MessageIdentifier::Heartbeat => 0xFD,
+            MessageIdentifier::StateTransitionRequest => 0xFE,
+            MessageIdentifier::StateTransitionCommand => 0xFF,
+            MessageIdentifier::Emergency => 0xFC,
```
TomLonergan03: why these numbers?

boards/stm32f767zi/src/tasks/can/board_heartbeat.rs
```diff
@@ -0,0 +1,99 @@
...
+                    from_board
+                );
+                emergency!(Reason::MissingHeartbeat);
+            }
+        }
+    }
+}
+
+/// Gives the boards a chance to wake up at the start.
+pub async fn wait_for_first_heartbeat(target_board: Board) -> Result<(), ()> {
```
TomLonergan03, line 68: :+1: 

was expecting to call this out as missing lol

boards/stm32f767zi/src/tasks/mqtt/send.rs
```diff
@@ -55,13 +57,21 @@ pub async fn mqtt_send_task(stack: &'static Stack<Ethernet<'static, ETH, Generic
...
     loop {
-        while !SEND_TO_MQTT_CHANNEL.is_empty() {
-            let message = SEND_TO_MQTT_CHANNEL.receive().await;
-            mqtt_client
-                .send_message(message.topic.as_str(), message.payload.as_bytes(), false)
-                .await;
-        }
-        Timer::after(Duration::from_millis(100)).await;
+        let message = MQTT_SEND.receive().await;
+        defmt::debug!("Sending MQTT message: {}", message);
```
TomLonergan03, line 71: are logs being sent over mqtt or just to serial? cos if you log each time you send a message over mqtt, then that log will trigger another log lol

boards/stm32f767zi/src/tasks/state_machine.rs
```diff
@@ -0,0 +1,69 @@
...
+            None => {
+                defmt::error!(
+                    "State transition failed. Invalid transition from {:?} to {:?}",
+                    state_machine.current_state,
+                    to_state
+                );
+            }
+        }
+
+        Timer::after(Duration::from_millis(10)).await;
```
TomLonergan03: why do you wait here as well as `await`ing the state transition request channel? i think i've seen it in a few different tasks

lib/communications/src/can_id.rs
```diff
@@ -0,0 +1,95 @@
...
+    fn from(val: CanId) -> Self {
+        let priority: u32 = if val.priority { 1 } else { 0 };
+        let board: u32 = u8::from(val.board) as u32;
+        let message_type: u32 = u8::from(val.message_data_type) as u32;
+        let message_identifier: u32 = u16::from(val.message_identifier) as u32;
+
+        // Make sure that measurement_identifier is 13 bits
+        assert!(message_identifier < (1 << 13));
+
+        // Format: priority (1 bit) | message_type (8 bits) | message_identifier (13 bits) | board (8 bits) = 29 bits
```
TomLonergan03: :+1:

TomLonergan03 (approved):
this pr is too big so i can't make any comments on correctness, it should really have been like 3 500 line ones imo. looks good in general but it's beyond my context limit for sure

boards/stm32f767zi/src/tasks/state_machine.rs
davidbeechey: yeah that's wrong (probs left over from before the `await` stuff was added)

boards/stm32f767zi/src/tasks/mqtt/send.rs
davidbeechey: Haven't thought too much about logs yet, will do that in another PR. I suspect:
- Debug logs will be local board only, so safe to use there
- Error, warn, info logs will be sent over CAN and then MQTT

lib/communications/src/identifier.rs
davidbeechey: iirc because they were just being ANDed to another id so this is just growing from both ends, but should probs be updated to be nicer than that

boards/stm32f767zi/src/tasks/mqtt/send.rs
davidbeechey: quite right, it is called just as an async function [here](https://github.com/Hyp-ed/hyped-2025/pull/94/files/8cb70be5e95e7a340192a7b4cabfca1c301b8278#diff-755a4ee53f890b07d43faaf522124876cd69d1f3833b7d0d9be5114759555140:~:text=///%20Split%20up%20the,%7D)

boards/stm32f767zi/src/tasks/mqtt/base_station_heartbeat.rs
```diff
@@ -0,0 +1,22 @@
...
+pub async fn base_station_heartbeat() {
+    loop {
+        MQTT_SEND
+            .send(MqttMessage::new(
+                MqttTopic::Heartbeat,
+                String::<512>::from_str("").unwrap(),
+            ))
+            .await;
+        debug!("Sent heartbeat message");
+        Timer::after(Duration::from_millis(100)).await;
```
Aux1r: doc says you're sending a message to the MQTT broker ever second but is this not every 0.1 seconds?

boards/stm32f767zi/src/bin/boards/telemetry.rs
```diff
@@ -0,0 +1,108 @@
...
+    let current_state_sender = CURRENT_STATE.sender();
+
+    loop {
+        // All main loops should have logic to handle an emergency signal...
+        if EMERGENCY.receiver().unwrap().get().await {
+            defmt::error!("Emergency signal received! Cleaning up...");
+            // ... and take appropriate action
+            current_state_sender.send(State::Emergency);
+            // Wait for the emergency signal to be sent
+            Timer::after(Duration::from_secs(1)).await;
```
Aux1r, line 104: Is there a reason that in some parts of this PR you're using `Duration::from_milliseconds` and in some you're using `from_secs`? Very picky but it would be good to standardise this and go with 1 or the other?

lib/core/src/config.rs
```diff
@@ -13,6 +16,11 @@ use config_to_rs::config_to_rs;
 #[config_to_rs(yaml)]
 pub struct Config;
 
+// TODOLater: this should be in a config
```
Aux1r, line 19: job for a new PR or add it to this one?

Aux1r (commented):
this is an absolute monolith of a PR, I don't fully understand what's going on in a half of it (skill issue). I left some nit picky comments but as most of this is beyond me, LGTM

boards/stm32f767zi/src/bin/boards/telemetry.rs
davidbeechey: Yeah this should be standardised, there shouldn't be a mix - particularly within loops they should all be frequencies I think because that makes the most sense

lib/core/src/config.rs
davidbeechey: definitely for another PR, my plan is to have 2 follow ups (by me or someone else):
- Move all constants to config
- Logging

PR #96, opened 2025-05-14
INFRA - Config

davidbeechey:
Moves lots of constants into YAML config files using the [`config_to_rs`](https://crates.io/crates/config_to_rs) crate.


boards/stm32f767zi/src/bin/boards/telemetry.rs
```diff
@@ -34,14 +35,13 @@ use hyped_boards_stm32f767zi::{
         network::net_task,
         state_machine::state_machine,
     },
-    telemetry_config::{BOARD_STATIC_ADDRESS, GATEWAY_IP},
 };
 use hyped_communications::boards::Board;
-use hyped_core::log_types::LogLevel;
+use hyped_core::{config::TELEMETRY_CONFIG, log_types::LogLevel};
```
TomLonergan03, line 40: i think config is an unused import

TomLonergan03 (approved):
v nice

boards/stm32f767zi/src/bin/boards/telemetry.rs
davidbeechey: looks like it, but it's used in the `set_up_network_stack!` macro so needs to be imported there

miksquared (approved):

H-Allen (approved):

jpfbastos (approved):
lgtm!

PR #99, opened 2025-06-11
SNS - LP BMS

H-Allen:


boards/stm32f767zi/src/tasks/sensors/read_lp_bms.rs
```diff
@@ -0,0 +1,97 @@
...
+                latest_bms_sender.send(Some(battery_data.clone()));
+
+                let board = *THIS_BOARD.get().await;
+
+                // Send all BMS information over CAN
+                can_sender
+                    .send(CanMessage::MeasurementReading(MeasurementReading::new(
+                        CanData::F32(battery_data.voltage),
+                        board,
+                        measurement_id,
```
davidbeechey, line 36: The measurement ID should be unique for each "thing" we're measuring, so all of these will need their own. You can add them to `/config/pods.yaml` and hard-code them in here using the `MeasurementId`. Otherwise it'll be hellish to pass into this task (and we'll probs only have one LP BMS so don't need to worry about that)

lib/sensors/src/lp_bms.rs
```diff
@@ -0,0 +1,117 @@
+/// Driver for the TinyBMS s516 30A Battery Management System using CAN.
+/// Used to monitor battery status and health.
```
davidbeechey, line 2: Clippy is complaining about these comments, so either put them somewhere else in the file or change them to `//` instead of `///` (probably the former)

lib/sensors/src/lp_bms.rs
```diff
@@ -0,0 +1,117 @@
...
+        Ok(BatteryData {
+            voltage,
+            current,
+            max_cell_mv,
+            min_cell_mv,
+            temperatures_c,
+            cell_voltages_mv,
+        })
+    }
+}
```
davidbeechey, line 117: Any reason this is in a separate `impl` block to the block above? Looks like they're the same

lib/sensors/src/lp_bms.rs
```diff
@@ -0,0 +1,117 @@
...
+
+
+pub struct Bms<'a, T: HypedCanTx + HypedCanRx + 'a> {
+    can: &'a mut T,
+}
+
+impl<'a, T: HypedCanTx + HypedCanRx> Bms<'a, T> {
+    const NODE_ID: u8 = 0x01;
+    const REQUEST_ID: u32 = 0x400 | Self::NODE_ID as u32;
```
davidbeechey, line 26: What is `Self::NODE_ID`?

lib/sensors/src/lp_bms.rs
```diff
@@ -0,0 +1,117 @@
...
+    const RESPONSE_ID: u32 = 0x500 | Self::NODE_ID as u32;
+
+    fn send_simple_request(&mut self, cmd: u8) -> Result<(), CanError> {
+        let frame = HypedCanFrame::new(Self::REQUEST_ID, [cmd, 0, 0, 0, 0, 0, 0, 0]);
```
davidbeechey, line 30: likewise what is `Self::REQUEST_ID`?

davidbeechey (changes requested):
Nice! A few things to look at, plus making clippy and the formatter happy. Also could you merge main back into this branch, you seem to have included a lot of accel code? Thanks!

